PR #25, opened 2012-01-30
IP address as additional authorization parameter. Hook API to track muc rooms and opened sessions.

rustamserg:
- IP address as last parameter for authorization modules. Modules which are not supported additional IP address parameter work without changes.
- pre_open_session hook to add custom check for incoming IP address and username before opening actual session
- mod_muc_hook module to have possibility to implement custom logic when room is opened and add message filtering

### DIFF
--- a/src/mod_muc/mod_muc_hook.erl
+++ b/src/mod_muc/mod_muc_hook.erl
@@ -1,0 +1,19 @@
+-module(mod_muc_hook).
+
+-export([room_state/3,
+		room_user/5,
+		room_message/5,
+		room_message_private/7]). 
+
+
+room_state(Host, Room, State) ->
+	ok.
+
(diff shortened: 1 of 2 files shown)


ph4r05:
Hi there!

This pull request seems to be 5 years old. Any changes to this? Are maintainers thinking about merging this? 

I would appreciate passing IP address to the external auth handler as well. In my opinion it would be great to separate this PR into two. First - the IP, the second - MUC. What do you think?

cromain:
indeed, it's BEST to avoid PR with multiple unrelated changes.
Please note Holger already worked on a patch for adding IP
https://gist.github.com/weiss/8f64dc5cb98f08487563

I'm closing this PR as we won't apply it "as this"
Also please note, as stated on #1731, the ACL way for handling IP checks is likely better than changing API.

PR #30, opened 2012-05-12
Add support for publish-only in hometree derived nodes

teisenbe:
I can port these changes to 3.x also, but it'd be a nice thing to have in the next 2.1.x release.


cromain:
fixed in 6e4d6ca

PR #33, opened 2012-06-26
2.1.x

altenwald:
Redis support for roster:

 * Basicly, the main files modified were:
   - configure / configure.ac
   - Makefile.in
   - eredis directory was added to src directory (to do easier the redis integration)
   - ejabberd_redis (to handle calls to redis)
   - mod_roster.erl (obviously  )

 * The config file could add the next string:

   {redis_server, {"localhost", 6379, 0, ""}}.

   The params means:
   - host
   - port
   - database
   - password

   The mod_roster module should use "db_type" to change from mnesia to redis database:
   {mod_roster,   [{db_type, redis}]},

 * To install the solution, the commands sequence should be:

   aclocal
   autoconf
   ./configure --prefix=/usr/local --enable-eredis
   make
   make install

 * Redis is used as schema throught keys, therefore we use this key to access to roster list (or set):
   usererver:roster

   And this key to access to individual register and field:
   usererver:roster:jid:field

   The params means:
   - user: the user owner of roster.
   - server: domain for the user owner of roster.
   - roster: as itself.
   - jid: the contact in roster.
   - field: a individual data for the contact in roster.

That's all, tough.

Regards.


cromain:
closing as 2.1 is discontinued

PR #166, opened 2014-04-14
Add support for XEP-0198: Stream Management

weiss:
This is a complete implementation of XEP-0198 (Stream Management) for c2s connections.  It adds four c2s options (all documented in `doc/guide.tex`):

```
stream_management: true|false (default: true)
resume_timeout: Seconds (default: 300)
resend_on_timeout: true|false (default: false)
max_ack_queue: Size (default: 500)
```

We've been using this patch in production for a while now, and it seems to work fine for us.  I tested it against various clients and corner cases.

I've also tried to keep the patch as non-intrusive as possible.  While it has to modify some of the existing code, a quick `git diff` might look worse than it is: There's lots of whitespace changes in `ejabberd_c2s:terminate/3` (as the existing code is wrapped into a new `case` statement); and the c2s `#state` is updated in some places, so the `#state` variable names were modified.

If you decide to merge this and any bugs pop up, I'd be happy to look into them.


nectarbits:
I can anyone help me with this:
http://stackoverflow.com/questions/33985984/xmpp-ejabberd-online-offline-status-issue-when-internet-connection-loss

weiss:
@nectarbits, I would give up the requirement of having the presence status of mobile devices updated instantly, as that's plain impossible with a TCP-based protocol.  You *could* use a very short ping interval of course, but that will drain your users' batteries.

You might want to read a [related FAQ entry][1].

[1]: https://www.ejabberd.im/faq/tcp

PR #391, opened 2014-12-28
Ability to install rich modules of ejabberd.

loguntsov:
All modules must placed to ebin directory via simple beam-files list.
Simple modules for ejabberd must placed with this files also.
Rich modules require any dependencies and this make problem with their installation.

This solution permits install modules with dependencies as direcory of ebin directory.
For example if the ejabberd installed to /lib/ejabberd, then any modules will can install to the /lib/ejabberd/ebin/my_module.

The beam-filles of module must placed to ebin directory of my_module directory (my_module/ebin)
And all beam-files of module dependencies must placed to ebin directory of deps directory (my_module/deps/my_dependence/ebin ).


cromain:
This PR is not needed with recent addition of modules management in ejabberd
see: https://blog.process-one.net/easy-installer-and-structure-for-ejabberd-contributed-modules/

loguntsov:
Please, explain me:
How can i use dependencies for rebarized-plugins ?

cromain:
rebarized plugins with dependencies are not supported for the beta version of ext_mod, but we can simply add support without changing ejabberdctl. indeed, ext_mod dynamically handle the code path on load. by now we just call code:add_patha(module_ebin_dir(Module)). but we should improve this to add all directories under module's deps directory first. Anyway the best solution is to improve ext_mod. So i reopen this ticket until i add the following changes to ext_mod:

cromain:
i'll add this in coming days, this is tracked on EJAB-1737 ticket.

cromain:
it won't be easy to handle rebar so i just added a minimalist support
Either contributed module include dependencies this way 
```
  deps/
    dep1/
      src/
      include/
    dep1/
      src/
      include/
```
Or includes rebar.config or rebar.config.script:
In this case, only git is supported (if git command available)
and ext_mod checkout code in deps directory.

In both case, only basic built procedure is supported. ext_mod
does not do more than bare compilation like this:
erlc -I include src/*erl

PR #397, opened 2015-01-08
Update README

toopay:
Add the `autogen.sh` execution to generate a new autoconf prior anything else.

### DIFF
--- a/README
+++ b/README
@@ -118,7 +118,7 @@ To compile ejabberd you need:
 ### 1. Compile and install on *nix systems
 
 To compile ejabberd execute the commands:
-
+    ./autogen.sh
     ./configure
     make
 


zinid:
autogen.sh is only needed when building from git

toopay:
@zinid I'm aware of that. But the current README doesn't explicitly said to : checkout to `2.1.x` branch's `src` prior execute `./configure`, so its pretty logical to assume that current build instruction is for master branch, no?

zinid:
I don't know, I thought this README is for tarball actually :)

weiss:
@toopay, what README are you talking about?  ejabberd's [current README file][1] doesn't mention the Git repository, as far as I can see.  I think developers who use the repo are expected to know what they're doing :-)

[1]: https://raw.githubusercontent.com/processone/ejabberd/ad4b41730e488c75b84894f2ff311f5409f32543/README

toopay:
> I think developers who use the repo are expected to know what they're doing

Fair enough.

badlop:
By the way, the Guide includes a small paragraph about Git, so I've added that comment there.

PR #412, opened 2015-01-21
Fix overflow in XEP-0203 delay: if microseconds exceeded 999499, *** was...

nbruning:
... put in the formatted timestamp


weiss:
Good catch, thanks a lot for spotting the issue and tracking it down.

There's a small bug in the patch: The [`Max` clause][1] of the `case` expression will never match, as `Max` is a `float()` value which is matched against an `integer()` there.

While at it, I guess we could [increment the number of seconds][2] instead of decrementing the fractional part in case of overflow, just to be slightly more correct/comprehensible.

Thanks again for this report (and sorry for the original bug!).

[1]: https://github.com/processone/ejabberd/pull/412/files#diff-f2c18ecff0d9ad9bf8e81d78cd6d07efR722
[2]: https://github.com/weiss/ejabberd/commit/4220e8dd24dac5bbe0b8ee3a4bb7fae7809b076e

PR #426, opened 2015-02-06
Migration of scram passwords with piefxis (riak and internal only)

hairyhum:
Current version is throwing `badxml` exception when trying to export password in SCRAM format. 
Added encoding/decoding password to binary format and handling non `iostring` passwords in `try_register` for `internal` and `riak` auth.


badlop:
Hairyhum, the "Hook on group message" commit is unrelated to this feature, right?

hairyhum:
Yes.

PR #435, opened 2015-02-16
Fixed node type.

saczawap:
Node type depends on children of its node. Following XEP-0060 node type
should matche type given in configuration.


cromain:
I've applied same change after pubsub refactor in b82ed1c

PR #443, opened 2015-02-25
Permit modifying affiliations via an event

benlangfeld:
Allows us to implement mod_muc_admin:set_room_addiliation properly


mremond:
Hi @benlangfeld,
Is it for solving your issue with MUC feature on #436 ? Do we still need to find a solution for that one ?

benlangfeld:
Partially, yes, @mremond, though @bklang wants to use peer-kick also since we run this ban job at low-priority in the background, and so it could take a few seconds, making the UX sluggish.

PR #448, opened 2015-02-27
update japanese translation.

hamano:
I've updated new translation.
BTW, Some XMPP client such as PSI just started to replace "Jabber" to "XMPP" in message.
I think It's about time replacing message for ejabberd.
Thank you.


badlop:
Thanks! Now Japanese translation is completed :)
The change in Jabber/XMPP could be considered with other changes in translations that people have mentioned last year.

Issue #456, opened 2015-03-04
mod_irc doesn't get restarted after a crash


ehuelsmann:
After a single failure (crash), the service becomes completely unavailable. Holger says the service isn't configured to be restarted. Maybe it's an idea to restart the service after a delay?

mremond:
The IRC components should be put in proper Erlang supervision tree. I am proposing this tasks as teaser tasks for Google Summer of code.

mremond:
There is a tentative patch attached on Pull Request #475

badlop:
mod_irc is not maintaned anymore. The code was moved to ejabberd-contrib in June (3a5d2dbed8f089b112eba7998e2ad0d2508f01cb), so that it doesn't get lost. If this problem is still present and somebody wants to fix it, a new ticket in [ejabberd-contrib](https://github.com/processone/ejabberd-contrib) can be created, referencing this old one.

PR #475, opened 2015-03-14
Issue #456: Attempt to auto restart mod_irc in case of crash

kansi:
changed mod_irc from temporary to transient child and placed it under ejabberd_mod_irc_sup supervisor.

### DIFF
--- a/src/mod_irc.erl
+++ b/src/mod_irc.erl
@@ -92,7 +92,7 @@ start(Host, Opts) ->
     Proc = gen_mod:get_module_proc(Host, ?PROCNAME),
     ChildSpec = {Proc, {?MODULE, start_link, [Host, Opts]},
 		  transient, 1000, worker, [?MODULE]},
-    supervisor:start_child(ejabberd_mod_irc_sup, ChildSpec).
+    supervisor:start_child(ejabberd_sup, ChildSpec).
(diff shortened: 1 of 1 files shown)


zinid:
ejabberd_mod_irc_sup is a `simple_one_for_one` supervisor. AFAIK, you cannot attach transient children to it, as it doesn't even keep any childspecs.

kansi:
yeah. looked at ejabberd_sup its one_for_one, I think it was correct previously to have mod_irc under ejabberd_sup but as transient child so that it can be restarted.

mremond:
@zinid How does it look, now ?

zinid:
@mremond more or less. But now I don't see any purposes of ejabberd_mod_irc_sup :)

mremond:
Isn't that mod_irc_sup for the connections themselves ?

badlop:
mod_irc is not maintaned anymore. The code was moved to ejabberd-contrib in June (3a5d2dbed8f089b112eba7998e2ad0d2508f01cb), so that it doesn't get lost. If this problem is still present and somebody wants to fix it, a new ticket in [ejabberd-contrib](https://github.com/processone/ejabberd-contrib) can be created, referencing this old one.

Issue #459, opened 2015-03-06
SQLite support for ejabberd


mremond:
Here are the needed steps:

- Integrate and existing driver / extension for SQLite in Erlang in the build and dependancy chain.
- Expand ejabberd_odbc.erl to support SQLite.
- Create SQLlite schema as lite.sql file.
- Example command line to generate SQLite empty database
- Provide empty sqlite file structure as a starting point to use ejabberd with SQLite

zinid:
There is no need to create a driver :)
There are several drivers readily available :)

mremond:
Thanks, I will update the description.

saa:
Hi. I start add sqlite3 support for ejabberd. I chose https://github.com/alexeyr/erlang-sqlite3, because this is driver.

And I have question about config for sqlite. Use odbc_server param for set database path? Example: odbc_server="test.db" or odbc_server='"/tmp/test.db" ?

mremond:
For SQLite we probably need only two parameters:

```
odbc_type: sqlite
odbc_database: "/path/to/database.db"
```

saa:
Hello. I added sqlite support and open pull request, but I have one question. Thank you.

mremond:
Hi,

We see you had trouble with your pull request. Did you finally manage to prepare it ?

saa:
Hi. Yes, the second time I hurried. Need fix create schema.

saa:
Hi. I made new PR. But have error (tests result) with timeout recv: timetrap_timeout.

mremond:
Thanks, we will have a look.

mremond:
Can you please point us to the pull request ?

saa:
https://github.com/processone/ejabberd/pull/486

jianAjian:
there is a error occured  when I did it .

jianAjian:
There is a error occured when I did it .
My steps are as follows:

1. edit the ejabberd.yml
 ```
 auth_method: sql
 
 ....
 sql_type: sqlite
 sql_database: "/Applications/ejabberd-16.06/database/ejabberd.db"

 ```

2.  create a db with lite.sql  then put it under the 'database'  

3. start

4. get a error
 
![image](https://cloud.githubusercontent.com/assets/12096759/16574808/83593702-42b5-11e6-8043-e7d2f9166ca3.png)

What did I miss ?

here are logs

```
2016-07-05 13:06:59.090 [info] <0.7.0> Application lager started on node ejabberd@localhost
[...]

PR #476, opened 2015-03-15
Add sqlite support

saa:
Hi.
I added support SQLite3. But I have one question: for this case "Provide empty sqlite file structure as a starting point to use ejabberd with SQLite" (from issue #459), also need install sql/lite.sql. Path $PREFIX/lib/ejabberd/priv/sql/lite.sql it's ok?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *~
 \#*#
 .#*
+.edts
 /Makefile
 /config.log
 /config.status
--- a/configure.ac
+++ b/configure.ac
@@ -17,8 +17,8 @@ fi
 
 # Checks Erlang runtime and compiler
 AC_ARG_WITH(erlang,
-	AC_HELP_STRING([--with-erlang=dir],
-		[search for erlang in dir]),
+    AC_HELP_STRING([--with-erlang=dir],
+        [search for erlang in dir]),
 [if test "$withval" = "yes" -o "$withval" = "no" -o "X$with_erlang" = "X"; then
     extra_erl_path=""
 else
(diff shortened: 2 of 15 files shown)


zinid:
Hi, seems like Travis has failed.
Also, what's up with Redis? Why did you remove it from the configure script?

zinid:
Another issue is indentation (in YAML file): please don't touch empty lines or spaces, it will produce conflicts on merging.

saa:
I forked before changes were with redis.
Thank you for review. I will fix and push again.

saa:
Maybe better create new pull request? Without changes in YAML with empty lines or spaces.

zinid:
Yes, I close this one

mremond:
@saa Thanks, Please send us a pull request that is clean for integration, it will help.

Regarding your question: Mnesia database is created in $PREFIX/var/lib/ejabberd/
We should probably create sqlite empty structure in:
$PREFIX/var/lib/ejabberd/ejabberd-sqlite/lite.sql

Note that if the file already exist, it should not be overwritten to avoid data loss, but should be created as lite.sql-new

@zinid does it make sense ?

saa:
@mremond copy sql schema to $PREFIX/var/lib/ejabberd/ejabberd-sqlite/lite.sql from sql/lite.sql?

I wanted copy lite.sql to priv dir $PREFIX/lib/ejabberd/priv/sql/lite.sql. Maybe I misunderstood?

mremond:
What does lite.sql contain ? Is it sqlite database schema or is it database file itself (but empty) ?

saa:
It's sqlite database schema

saa:
And about: "Provide empty sqlite file structure as a starting point to use ejabberd with SQLite", at now I do it when init ejabberd_odbc_sup.

PR #486, opened 2015-03-19
Add SQLite support

saa:


zinid:
Did you run the tests locally? Why don't they work?

saa:
Yes, I ran locally. And get error with connection timeout for 4 tests.

saa:
@zinid I found reason and I'll fix it and push.

zinid:
Awesome!

saa:
Strange. make xref failed, local is ok.

zinid:
Most likely you're using some relatively new functions introduced in recent Erlang versions. Such as erlang:binary_to_integer/1,2.

zinid:
Regarding your last commit: we typically do not use such coding approach, i.e. we're not using guard functions such as is_list, is_integer and so on for type verification: this should be done by Dialyzer. Also, binary_to_integer(5) breaks the function semantics.
Why do you need such tricks?

saa:
I don't use binary_to_integer/1,2.

And about trick: sqlite3 return integer when is integer. In ejabberd code use jlib:binary_to_integer for value return from db. So I use this trick.

zinid:
If you want this hack, then use it in a local code, do not spread this on all code via jlib.erl. Such `defensive code` leads to hard-to-track errors.

saa:
Ok. I'll fix it.

mremond:
@saa Hi, how is the fix going ?

saa:
@mremond Hi. Sorry, I digress.

@zinid It's ok? No need to fix?

zinid:
I fixed and improved everything

saa:
@zinid you are cool. Thank you.

PR #489, opened 2015-03-20
export CONTRIB_MODULES_PATH

sezuan:
Should be exported to be usable in ext_mod.


cromain:
you're right, i forgot this export. this is fixed now. (i don't know why auto-merge failed so i manually pushed fix) Thanks !

PR #491, opened 2015-03-23
Hook on group message

hairyhum:


weiss:
1. I think for some use cases it might make sense to also hand over the `#state.config` of the room.  (Yes, the room process could be [queried][1], but that's not free.)

2. Apart from that, I think the callback(s) should be able to modify or drop the message (which means the hook would have to be moved [out of][2] the `add_message_to_history/4` function).

3. Last not least, I'd use a different name for the hook (e.g., `muc_filter_message`, or just `muc_message` if you disagree with my previous suggestion).

[1]: https://github.com/processone/ejabberd/blob/b6999b126a287ddbe43c9f618356d12b51ccae03/src/mod_muc_room.erl#L741
[2]: https://github.com/processone/ejabberd/blob/b6999b126a287ddbe43c9f618356d12b51ccae03/src/mod_muc_room.erl#L960

mremond:
I totally agree that such a hook should be a filter. This is a good point.

hairyhum:
In this case should user see some errors about filtered out message and should message be added to history?

weiss:
> In this case should user see some errors about filtered out message

As far as `mod_muc_room` is concerned: No.  (The module that drops a message could generate error notifications if that's desired.)

> should message be added to history?

A dropped message should not be added to the history.

hairyhum:
There are error in build on `xref` step, can someone say what could cause it?

mremond:
Sometime Xref fails if code is targeting a specific version of Erlang.

mremond:
This is due to integration with mod_admin_extra.erl which is only partially finished. It should not be an issue with your code.

hairyhum:
Is everything OK with this pull request? Can you merge it?

weiss:
> Is everything OK with this pull request? Can you merge it?

Well, I raised some concerns [above][1].  I can suggest an alternative patch tomorrow or so.

[1]: https://github.com/processone/ejabberd/pull/491#issuecomment-85072085

mremond:
Thanks, @weiss 
I assigned you the ticket :)

weiss:
I've submitted a suggestion in PR #618.  It adds a `muc_filter_packet` hook which allows for modifying or dropping the message.  The callback receives the message, the room `#state`, the room JID, the user JID, and the nick name as arguments.  Everyone happy with that?

PR #503, opened 2015-04-01
Fix EJABBERD_OPTS quoting for Elixir

weiss:
This fixes an error message that pops up during e.g. `ejabberdctl iexlive` startup when `log_rotate_date` is set to `""`.


mremond:
@cromain The change seems good to me, but do you think it could break something regarding some specific shell ?

PR #518, opened 2015-04-07
fix otp version check in ejabberd_listener.erl listen_tcp/5

liudanking:
Starting from otp 17.0, function erlang:system_info(otp_release) only returns the main version.


weiss:
> Starting from otp 17.0, function erlang:system_info(otp_release) only returns the main version.

Good catch, thank you!

Note that `EVsn >= "R13B"; EVsn >= "17"` would match any OTP version, as e.g. `"R12" >= "17"` is `true`.  However, these days we don't support OTP versions older than R16 anyway, so I just removed the check.

liudanking:
Agree.

PR #524, opened 2015-04-09
Use SASL authzid as client identity if auth module permits it

benlangfeld:
This allows the authentication modules to perform SASL proxy authentication. It puts the onus on them to authorize the authcid to masquerade as the authzid. Doesn't currently implement such functionality in existing auth modules, since they cannot currently codify a relationship between the two identities. Does not permit the authzid to use a domain differently from the one of the connection.

Note: digest might not work, but I have no interest in it, being deprecated.

PR #525, opened 2015-04-09
MojoAuth Module

benlangfeld:
Permits use of MojoAuth (http://mojoauth.mojolingo.com/) in ejabberd. MojoAuth is a set of standard approaches to cross-app authentication based on HMAC which is specified in RFC2104.


benlangfeld:
This should be good to merge along with #524. Would it be possible to include in 15.05?

benlangfeld:
Is there any chance this might be reviewed for inclusion?

benlangfeld:
@badlop I really need this in several of my apps and running on an ejabberd fork is becoming tricky. Can this and #524 please be considered?

src/cyrsasl.erl
```diff
@@ -128,7 +128,7 @@ register_mechanism(Mechanism, Module, PasswordType) ->
 %%    end.
 
 check_credentials(_State, Props) ->
-    User = proplists:get_value(username, Props, <<>>),
+    User = proplists:get_value(authzid, Props, <<>>),
```
mremond: Changing parameters name: Doesn't it break other form of authentication ?

benlangfeld:
> We had never heard of MojoAuthentication before.

Even if the PR this is based on could be merged, that would be something. That PR does not introduce MojoAuth, but simply makes ejabberd more SASL compatible by allowing authentication modules to differentiate between authcid and authzid.

src/cyrsasl.erl
benlangfeld: Not really. These parameters are normally the same.

benlangfeld:
Rebased for an eventual merge. In the meantime I'll prep a PR to get this into ejabberd-contrib until such time as it's eligible for core.

mremond:
@benlangfeld Maybe this helps ?
https://github.com/processone/ejabberd/commit/46568fb9596d8203c635477b30c91d2928847dea

benlangfeld:
I saw that, thank you to everyone involved in getting that merged @mremond :)

Ashaman-:
@benlangfeld Is there any ETA on MojoAuth? I'm sure I'm not the only one that hates having to choose between SIP/TURN authentication and SCRAM passwords...

Neustradamus:
@benlangfeld Always possible for 18.12+?

benlangfeld:
I don’t have a need for this right now, but if anyone else has the time to push it across the finish line, be my guest.

cromain:
If needed for faster inclusion, it can be applied to ejabberd-contrib instead.

Neustradamus:
@badlop: It can be added in ejabberd-contrib?

badlop:
> It can be added in ejabberd-contrib?

If it works correctly with recent ejabberd, yes, it can be added.

Issue #535, opened 2015-04-17
Add whitelist option to mod_fail2ban.erl


schwindp:
Please check whether it's possible to add the option of whitelisting IPs that won't be affected by mod_fail2ban.

PR #536, opened 2015-04-17
Add mod_fail2ban option to whitelist IP addresses

weiss:
This adds a `whitelist` option for `mod_fail2ban` as suggested in issue #535.


mremond:
Thanks!

zinid:
1) Why not use IP address ACL?
2) We do not support guide.tex anymore, as everything is now moved to http://docs.ejabberd.im/

weiss:
> Why not use IP address ACL?

Good idea, I've now added an `access` option to specify a rule, instead.

> We do not support guide.tex anymore

Okay.

PR #543, opened 2015-04-21
Updated purge offline filter

dwarburt:
When the pubsub nodes publish_model is open, then anyone could have
items that need to be purged, even if the person going offline is
not affiliated with the node.  The previous guard expression aborts
purge offline incorrectly in this case.  Instead, continue with the
purge if the user is affiliated with the node or if the node has an
open publish_model.


cromain:
dwarburt is right, and the patch is good, but the entire function can be improved. i'm on it.

PR #561, opened 2015-05-05
Tag private MUC messages with muc#user namespace

weiss:
Make it possible for clients to identify private MUC messages.  One use case would be to filter out undesired carbon copies of those.

This has been suggested [on the standards list][1] by the authors of [Yaxim][2] and [Conversations][3].

[1]: http://mail.jabber.org/pipermail/standards/2015-May/thread.html#29819
[2]: http://mail.jabber.org/pipermail/standards/2015-May/029819.html
[3]: http://mail.jabber.org/pipermail/standards/2015-May/029821.html


mremond:
Hello @weiss, It looks like a good idea to me.

PR #573, opened 2015-05-18
Fix some string/binary issues with non-native ODBC drivers

sezuan:
1. Configure the odbc connection to return values as binary.
2. Change table names to binary.

### DIFF
--- a/src/ejabberd_odbc.erl
+++ b/src/ejabberd_odbc.erl
@@ -510,7 +510,8 @@ abort_on_driver_error(Reply, From) ->
 %% Open an ODBC database connection
 odbc_connect(SQLServer) ->
     ejabberd:start_app(odbc),
-    odbc:connect(binary_to_list(SQLServer), [{scrollable_cursors, off}]).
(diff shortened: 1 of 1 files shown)


badlop:
Hi Sezuan. Do you remember some cases affected by the problems you fixed? I mean, can you provide some example use that was problematic without your commit? Thanks

sezuan:
Sure, creating an account was not possible. As far as I can tell, the functions expect that table names and values are returned as binaries. For instance:

1. For the column names in https://github.com/processone/ejabberd/blob/master/src/ejabberd_auth_odbc.erl#L79 the column names are expected to be binaries, but the column names are returned as strings, see [col_name() = string() - Name of column in the result set](http://www.erlang.org/doc/man/odbc.html#id56920)
2. For the returned values: A few lines later the StoredKey/ServerKey is used in the #scram record. As far as I can see, this must be binary, but the odbc module return strings unless the binary option is set, see [data type mapping](http://www.erlang.org/doc/apps/odbc/databases.html#type)

I've tested this with NuoDB, but I assume this should not depend on the database.

Issue #578, opened 2015-05-25
An auth external module requires ext_mod to be started firsst


badlop:
If I install the ejabberd_auth_http module with:
```
$ ejabberdctl module_install ejabberd_auth_http
```

and enable it with
```
auth_method: http
```

then, I get this error when ejabberd starts:
```
11:23:33.577 [error] CRASH REPORT
 Process <0.39.0> with 0 neighbours exited with reason:
 call to undefined function ejabberd_auth_http:start(<<"localhost">>)
 in application_master:init/4 line 133
11:23:33.578 [info] Application ejabberd exited with reason:
 call to undefined function ejabberd_auth_http:start(<<"localhost">>)
```

I found this workaround:
```
--- a/src/ejabberd_app.erl
+++ b/src/ejabberd_app.erl
@@ -59,13 +59,13 @@ start(normal, _Args) ->
     ejabberd_rdbms:start(),
     ejabberd_riak_sup:start(),
     ejabberd_sm:start(),
-    ejabberd_auth:start(),
     cyrsasl:start(),
     % Profiling
     %ejabberd_debug:eprof_start(),
     %ejabberd_debug:fprof_start(),
     maybe_add_nameservers(),
     ext_mod:start(),
+    ejabberd_auth:start(),
     start_modules(),
     ejabberd_listener:start_listeners(),
     ?INFO_MSG("ejabberd ~s is started in the node ~p", [?VERSION, node()]),  
```

But I wonder if this is a proper solution.

zinid:
The patch looks fine to me.

zinid:
You can try to create a PR to check if Travis is ok with it if you don't wanna run the tests locally.

PR #582, opened 2015-05-27
An auth external module requires ext_mod to be started first (#578)

badlop:
Proposed fix for https://github.com/processone/ejabberd/issues/578


mremond:
Thanks :+1:

PR #642, opened 2015-07-06
Also send PEP notifications when local contact updates CAPS

weiss:
When a contact of a local user becomes available, the contact should receive any last PEP items of the local user the contact is interested in. We currently use two different hooks for this:

- `presence_probe_hook`  
This hook is executed when a contact becomes available.  `mod_pubsub` only [uses it][1] when a *local* contact goes online.  For remote contacts, it [is ignored][2].

- `caps_update`  
This hook is executed when the local user receives a new [CAPS hash][3] from a contact.  This (usually) happens when a contact becomes available, *and also when the local user becomes available*. (Because in both cases, the contact will usually send a presence packet with such a hash.)  `mod_pubsub` only [uses it][4] for *remote* contacts.  For local contacts, it [is ignored][5].

So, for *local* contacts, last PEP items are only generated when the contact becomes available.  For *remote* contacts, they are also generated when the local user becomes available.  The latter behavior would usually be undesired, but it makes sense as a workaround to the issue that setting `ignore_pep_from_offline: false` doesn't work for remote contacts: They will never receive last items from local *offline* users.  To mitigate this problem a bit, they will also receive last items when the local user goes online again.  This is the reason to use the `caps_update` hook instead of the `presence_probe_hook` for remote users.

As a side effect of using the `caps_update` hook, remote clients might also receive the last PEP items if they just update their CAPS hash (because the user enabled PEP, for example).  This did *not* work for local contacts.

This PR changes things so that local users also receive last PEP items if they send a new CAPS hash, and it clarifies the code a bit.

[1]: https://github.com/processone/ejabberd/blob/0770252e9bb773a041dc3853b92c6026bdcd46be/src/mod_pubsub.erl#L699
[2]: https://github.com/processone/ejabberd/blob/0770252e9bb773a041dc3853b92c6026bdcd46be/src/mod_pubsub.erl#L701 
[3]: http://xmpp.org/extensions/xep-0115.html
[4]: https://github.com/processone/ejabberd/blob/0770252e9bb773a041dc3853b92c6026bdcd46be/src/mod_pubsub.erl#L686
[5]: https://github.com/processone/ejabberd/blob/0770252e9bb773a041dc3853b92c6026bdcd46be/src/mod_pubsub.erl#L689

### DIFF
--- a/src/mod_caps.erl
+++ b/src/mod_caps.erl
@@ -243,27 +243,24 @@ c2s_presence_in(C2SState,
 		  error -> gb_trees:empty()
 		end,
 	   Caps = read_caps(Els),
-	   {CapsUpdated, NewRs} = case Caps of
-				    nothing when Insert == true -> {false, Rs};
-				    _ when Insert == true ->
-					case gb_trees:lookup(LFrom, Rs) of
-					  {value, Caps} -> {false, Rs};
-					  none ->
-					      {true,
-					       gb_trees:insert(LFrom, Caps,
-							       Rs)};
-					  _ ->
-					      {true,
-					       gb_trees:update(LFrom, Caps, Rs)}
-					end;
-				    _ -> {false, gb_trees:delete_any(LFrom, Rs)}
-				  end,
-	   if CapsUpdated ->
-		  ejabberd_hooks:run(caps_update, To#jid.lserver,
-				     [From, To,
-                                      get_features(To#jid.lserver, Caps)]);
-	      true -> ok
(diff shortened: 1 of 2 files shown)


cromain:
yes, the last item use case can not work with federated pubsub and/or remote users, so i had to hack a bit with caps_update, while involved some changes in mod_caps also.
your point is good, and the patch seems good, I'll check it.

cromain:
thanks !

PR #654, opened 2015-07-12
Remove --enable-nif flag

weiss:
Specifying `--enable-nif` or `--disable-nif` when running ejabberd's configure script has no effect anymore: NIF support is enabled by default and can only be disabled by building [the p1_xml dependency][1] with `--disable-nif`.  So, this commit removes the flag.

[1]: https://github.com/processone/xml


prefiks:
Merged in b5cc4a3a7d59136022bcfeca4dda7fd11344d4ca, Thanks!

Issue #367, opened 2014-11-26
RFC 6121 section 2.1.6 Roster Push - Server sending full JID on 'from'


mktoast:
I've just started digging in XMPP and ejabberd, developing an Android client using Smack.

I'm using shared rosters to connect each group's user with everyone else. If I request from the client a user roster, everything works as expected, retrieving the user's contact list, some of them from the shared rosters.

I've encountered the next problem: If I add a user to a shared roster from the admin panel, the client does get notified, but ignores it, following rules by RFC 6121 2.1.6 "Roster Push", because the 'from' attribute received is the Full JID, including the /resource:

- A receiving client MUST ignore the stanza unless it has no 'from'
  attribute (i.e., implicitly from the bare JID of the user's
  account) or it has a 'from' attribute whose value matches the
  user's bare JID <user@domainpart>.

Is this an ejabberd bug? Will this be patched?

Thanks in advance

badlop:
Hi, something like this solves the problem, right?
```
--- a/src/mod_shared_roster.erl
+++ b/src/mod_shared_roster.erl
@@ -1028,7 +1028,7 @@ push_item(User, Server, Item) ->
                                            children = [item_to_xml(Item)]}]}),
     lists:foreach(fun (Resource) ->
                          JID = jlib:make_jid(User, Server, Resource),
-                         ejabberd_router:route(JID, JID, Stanza)
+                         ejabberd_router:route(jlib:jid_remove_resource(JID), JID, Stanza)
                  end,
                  ejabberd_sm:get_user_resources(User, Server)).
 
```

McPo:
Is there an update regarding this issue?
Having to patch Smack each time has become tedious.

PR #711, opened 2015-08-24
Don't send full JID in 'from' of Roster Push (RFC 6121 section 2.1.6)

gardenia:
just providing @badlop patch from #367 as a pull request

I can vouch for this patch as I have been running with it for several months against various XMPP client implementations without issue.  It solves the problem with the Smack library which is the pretext for #367


gardenia:
tests were previously failing on this but it appears to be because my fork was out of date with upstream.  all tests are passing now.

gardenia:
going to close this and resubmit as a I messed up the brancing.

PR #721, opened 2015-08-28
get_room_options will not crash in case of query for a non-existent room

oxpa:
Room options should return an empty list in case of a non-existent room. 
This lets other modules check muc options with proplists:get_bool(), which is nice.


zinid:
This leads to an API change because function get_room_options/2. Some third-party modules may rely on `room_not_found` return value (for instance, to render "no such room"). I don't think the change is acceptable.

zinid:
I misread the code, the function get_room_options/2 passes processing to get_room_options/1, so doesn't return `room_not_found`, so the patch is acceptable.

PR #735, opened 2015-09-16
Context support into translations

oxpa:
There is a problem with current translations system: there is no way to distinguish between similar words used in different contexts. For example, the problem I faced, English doesn't have nouns declention, so there is no way to distinguish how to translate months names. Current translation uses genetives of names while I need nominative.

With my patch it is possible to specify a context for translation using `translate:translate` and .msg files.
The change is fully backward compatible, no need to change anything. While new translations may have .msg files with three fields in a tuple: `{Original text, Translation text, Context}`. `Context` is simply a text to indicate current use.

If this patch will be merged, I'll have several more minor fixes into current code using `translate` function.


badlop:
How would that feature work with the gettext-format files? I mean, do you know if PO files support the "context"?

oxpa:
Gettext support "context" itself, but I can't find it in ejabberd sources.
For PO file the keyword is msgctxt. I think,I'm able to patch contrib
script to produce translations PO with contexts

How would that feature work with the gettext-format files? I mean, do you
know if PO files support the "context"?

—
Reply to this email directly or view it on GitHub
<https://github.com/processone/ejabberd/pull/735#issuecomment-140780417>.

badlop:
Right, I mean to update extract_translations so it converts between MSGS <---> PO files keeping the context.

Neustradamus:
Always possible for 18.12+?

cromain:
@badlop what do you think about this PR ?

oxpa:
Guys, may we close this one? I needed that for a web related feature which is heavily outdated and was implemented the other way around.
At this point (3 years away) I have almost zero chances to have this completed or updated.

PR #739, opened 2015-09-21
Web mam

oxpa:
This patch provides web representation for MAM logs.
This requires erlydtl (https://github.com/erlydtl/erlydtl) as a dependency.
#735 is also highly desired, so I will update that pull request: it allows web templates to be translated better

Templates themselves are a subject for a change: I used previously used styles, so they are simple.

There are several important notes:
Assume, conference host is "conferences.domain.tld"
This patch assumes, logs are either requested directly from 'server.tld' domain or "Muc-Host" header contains appropriate host name. If these conditions are false, first virtual host MUCs are used.
This patch doesn't handle 'presence' messages, because there are still no "presence" stanzas stored in MAM DB.
This patch used with mnesia DB may cause pretty high load while composing a list of dates for logs. So a caching web server is really desired.
Theyre are 4 templates in total: 
1. a list of conferences
2. a list of years-months with logs for that conference
3. a list of day in month
4. a day log template

All of them are available under `/conferences` url.

There are two options for this module: 
`template_dir` - a directory with templates 
`jslinkify` - if a text in a resulting page similar to a link is transformed into a link using a js script in the page (or in erlang either way).

Things I'm going to implement later:
- custom 'Muc-Host' header
- better template layout and styling
- presence messages support (as they are implemented in mod_mam)

You can find a functional example at  http://oxpa.acnet.ru/logs/conferences/


Neustradamus:
Always possible for 18.12+?

PR #747, opened 2015-09-24
degenderise error message

kennethgillen:
Changing "he sent" to "they sent" - no reason for this to be gender-specific.


weiss:
> Changing "he sent" to "they sent" - no reason for this to be gender-specific.

Might be harder to grasp for non-native speakers, though.  I for one wasn't aware that this usage of "they" is common, and Wikipedia says the "[acceptance varies][1]".

[1]: https://en.wikipedia.org/wiki/Singular_they

kennethgillen:
The same article also states that "Its use in formal English has increased in recent times with the trend toward gender-inclusive language, but it has been used by respected writers for centuries.". 

As a native speaker I can't tell if this really will make things more difficult to understand - it's certainly normal, everyday language for me.

kennethgillen:
Would "This participant is kicked from the room because an error message was sent by them" be clearer?

weiss:
> "Its use in formal English has increased in recent times with the trend toward gender-inclusive language, but it has been used by respected writers for centuries."

I guess I should read more books of "respected writers" :-)

> Would "This participant is kicked from the room because an error message was sent by them" be clearer?

Personally I still stumble over this while reading the message, simply because it's unfamiliar to me.  But I should probably just get accustomed to the singular they.

zinid:
I just checked 3 English-Russian dictionaries and none of them mention "singular they" :)

zinid:
As a technical note: the patch is incomplete, because none of 26 translated files was changed accordingly. I think in all those files the original string should be changed (without touching the translated one).

kennethgillen:
@zinid: thanks, good point. Original text in the translations updated now, too. (Even if the team don't want to update the text.)

badlop:
The proposed change solves the problem only for English native speakers and other languages that have an equivalent word.

I propose to rewrite the sentence in a way that does not require to provide the subject two times, so the problem is evaded completely.

For example, the original:
> This participant is kicked from the room because he sent an error presence.

could be replaced to something like:
> This participant sent an error presence, and consequently gets kicked from the room.

zinid:
@badlop a translator of the corresponding language may change this in his translation. I don't think this should affect all translated strings. So I vote for the solution proposed by @kennethgillen. It's simple and doesn't require a call for translators (which is always pain).

weiss:
Can't we use the the sentence proposed by @badlop without touching the translations?

zinid:
@weiss well, yes, but the sentence is a bit cumbersome for me.

zinid:
Anyway, we need to fix the issue as this is not the first time it arises (the previous one has been rejected).

pwalczysko:
How about replacing ....he sent... with he/she sent... ?

zinid:
@pwalczysko not sure if "he/she" is better than "they". I would accept what native speakers suggest. And for them it looks like "they" is better.

badlop:
I also agree to accept the "Singular they" solution for now, but considered only as temporary: it doesn't solve the problem for non-natives, and doesn't solve it for the other 24 languages. Basically, this change only solves the problem for a very small part of ejabberd users.

Once committed I'll open another ticket for a future ejabberd version in order to rewrite the sentence to not have any pronoun at all.

badlop:
It is possible to produce a string like this with a small patch:
> ~s sent an error message: ~s. Consequently, ~s gets kicked from the room.

This one is easy to translate to other languages and to understand by anybody with basic English knowledge.

A calling to translators for next 15.10 release is required anyway, because many missed last release and there are new strings.

zinid:
For the record, it is impossible to translate gender agnostically the phrase "gets kicked" in Russian. Because verbs in Russian have genders too (which depends on the corresponding nouns).
So this will not work as well.

badlop:
Oh, and other languages may have the same problem with that sentence. So it appears the most cross-language translatable string is one that doesn't specify the user name. My first attempt looked cumbersome, but maybe it can be improved, what about this?

> It is not allowed to send error messages to the room. This participant (~s) sent an error presence (~s) and gets kicked from the room.

cromain:
@badlop +1 for your latest suggestion. but this would require work on all translation right ?

PR #748, opened 2015-09-24
Remove bashisms in {join,leave}cluster scripts

joudinet:
So they can run in any POSIX shell, not bash only.

Bashisms have been detected with the checkbashisms program from the
devscripts package:
$ checkbashisms tools/*cluster
possible bashism in tools/joincluster line 18 ('function' is useless):
function error
possible bashism in tools/joincluster line 70 (sourced script with arguments):
. $CTL stop 2>/dev/null >/dev/null
possible bashism in tools/leavecluster line 14 ('function' is useless):
function error
possible bashism in tools/leavecluster line 48 (sourced script with arguments):
. $CTL stop 2>/dev/null >/dev/null

### DIFF
--- a/tools/joincluster
+++ b/tools/joincluster
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Add the current ejabberd node in a cluster
 
@@ -15,7 +15,7 @@
 # 30 : network issue
 # 31 : node names incompatibility
 
-function error
+error()
 {
     echo "Error: $1" >&2
     exit $2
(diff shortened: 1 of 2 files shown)


cromain:
your patch breaks the script
Error: can't find ejabberdctl

this is due to lack of ERL value.
the script called . $CTL .... to keep values initialised in the ejabberdctl context. from here we need ERL and ERLANG_NODE
removing local evaluation of CTL just make cluster script unable to plug on configured ejabberd

badlop:
Right Chris, adding "exec" solves the calling problem without adding any bashism. I'll commit that.

cromain:
good to me now

joudinet:
Ok, another possibility (to avoid the cost of exec) is to source the joincluster and leavecluster script from the CTL script and use the set built-in just before sourcing to set the script arguments.
For example:
```
join_cluster()
{
  set -- $*
  . "$EJABBERD_BIN_PATH"/joincluster
}
```

badlop:
As noted by zhchbin in https://github.com/processone/ejabberd/commit/deccf91280f3ed47719f965f03f0122297f51848, using 'exec' introduces another problem: that line is the last called in the script, so it doesn't work correctly.

PR #781, opened 2015-10-15
Fix-up mix.exs; Include mix.lock

scrogson:
## Remove `:ehyperloglog`

`ehyperloglog` dependency was removed in 83dd79a but was not removed from the list of `included_applications`.

## Add `mix.lock`

From the documentation on [hex]:

> `$ mix deps.get` will fetch dependencies that were not already fetched. Dependency fetching is 
repeatable, Mix will lock the version of a dependency in the lockfile to ensure that all developers will get the same version (always commit `mix.lock` to version control). `$ mix deps.update` will update the dependency and write the updated version to the lockfile.

[hex]: https://hex.pm/docs/usage


mremond:
Thanks :+1:

PR #790, opened 2015-10-21
Configurable ping ack timeout for mod_ping 

gardenia:
this patch allows the ping ack timeout (the length of time taken to deem that a ping has not being responded to) to be configurable. it still defaults to 32 seconds as per the constant IQ_TIMEOUT (in ejabberd_local.erl) this just makes it optionally configurable (in my case I needed to be more trigger happy than 32 seconds)

### DIFF
--- a/src/mod_ping.erl
+++ b/src/mod_ping.erl
@@ -63,6 +63,7 @@
 	{host = <<"">>,
          send_pings = ?DEFAULT_SEND_PINGS :: boolean(),
 	 ping_interval = ?DEFAULT_PING_INTERVAL :: non_neg_integer(),
+	 ping_ack_timeout = undefined :: non_neg_integer(),
 	 timeout_action = none :: none | kill,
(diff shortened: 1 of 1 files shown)


zinid:
On line 66: there is no need to write `undefined`. Just write:
```
ping_ack_timeout :: non_neg_integer(),
```
On line 112: the default value for get_opt is `undefined`, so no need to write it there as well.

Other stuff is ok. Please recreate the PR with a single commit and we will merge it (there are two many commits for such a simple patch).

zinid:
Ah, also please create a PR for https://github.com/processone/docs.ejabberd.im describing new option (you need to improve `mod_ping` module description in `content/admin/guide/configuration.md`).

gardenia:
thanks.  updated this PR as requested.  

PR for docs here: https://github.com/processone/docs.ejabberd.im/pull/6

PR #795, opened 2015-10-26
port of the structure of the ejabberd to the Oracle Database

softctrl:
I use this script to create the database using Oracle 11g


zinid:
Wow, cool
How do you manage oracle sql connection? Via odbc driver?

softctrl:
i make the access using odbc, now i am make tests to see if exists any
problem with the modules that i need here.
about the script maybe some fields have a truncate exception because i use
varchar2 (max limit is 4000), but like a say i am making tests to validate.

zinid:
Good. Keep us updated.

softctrl:
Hi, today i found many problems using oracle, i start to refactoring the project because i need to use oracle (the client wish), so, when i finish i let you know.

softctrl:
i am making some tests with riak

Issue #778, opened 2015-10-14
Request: "Stable" branch


SafwatHalaby:
Currently, the only way to obtain stable releases is through downloading tarballs from the ProcessOne site.

Having a stable branch would be more convenient. It also comes with incremental compilations out of the box: When a previous version was already compiled, the next `make` command takes considerably less time.

weiss:
> Currently, the only way to obtain stable releases is through downloading tarballs from the ProcessOne site.

Well, you could checkout the release tags, e.g. `git checkout 15.0`, and build those.

SafwatHalaby:
Oh, sorry.

prefiks:
Unfortunately i don't think we pin dependences before tagging, so you still would get floating deps when checking out tag, but not when downloading tarball.

SafwatHalaby:
I see. What about fetching and compiling `master` branch just after a release is tagged? I guess it would provide something nearly identical to the tarballs, wouldn't it?

weiss:
> Unfortunately i don't think we pin dependences before tagging, so you still would get floating deps when checking out tag, but not when downloading tarball.

Ah, indeed.

And you must also take care the other way round: If the dependencies are downloaded already, just rerunning `make` won't update them (but `make update all` will).

weiss:
> What about fetching and compiling master branch just after a release is tagged? I guess it would provide something nearly identical to the tarballs, wouldn't it?

Yes.  And you can run `tools/set-dep-versions` at that point to pin the dependencies to their current versions.

SafwatHalaby:
`make update all` doesn't seem to be a valid command.

weiss:
> `make update all` doesn't seem to be a valid command.

What happens if you run it?

SafwatHalaby:
Sorry, I invoked `make update all` before ever doing a `make`.
```
git clone https://github.com/processone/ejabberd.git && cd ejabberd && ./autogen.sh && ./configure && make update all
...
Makefile:96: recipe for target 'update' failed
make: *** [update] Error 2
```
it all works fine if I `make` before `make update all`.

So, in summary, is this the right way to obtain a stable branch from the repo?:
```
1. Fetch the master branch just after a `release` is tagged
2. autogen
3. make update all
4. make
```

My make skills are unimpressive, so I may have got it wrong.

weiss:
Ah right, `make update` won't download the dependencies if they aren't there yet.  `make deps update all` would work if you need a single command line that also works for this case.

[...]

PR #806, opened 2015-10-30
Added instructions to install from release tag

SafwatHalaby:
Please review the comments at #778 before accepting this PR.

### DIFF
--- a/README
+++ b/README
@@ -124,6 +124,16 @@ necessary if your source tree didn't come with a `configure` script.
     ./configure
     make
 
+If using git, the above sequence compiles the latest master. To compile a
+specific stable version of ejabberd from git, execute the following commands.
+This is irrelevant if you are installing from a tarball.
+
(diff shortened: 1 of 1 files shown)


weiss:
Apart from [my confusion][1] as to why you're using the `git-submodule(1)` command, things like checking out release tags aren't specific to the ejabberd repository but rather just generic Git knowledge.  Personally I think it would be good to not start turning the `README` file into a Git tutorial, because if you try to cover all possible use cases people might be interested in, that will add complexity and make the docs more confusing.

I think the `README` file is still primarily meant to document building from the source tarball, and those who prefer to use the Git repository (presumably for development) are supposed to know the necessary Git basics (or to know where to find the [relevant docs][2]).

But that's just my 2¢, it's not my decision :-)

[1]: https://github.com/processone/ejabberd/issues/778#issuecomment-152645084
[2]: https://git-scm.com/documentation

PR #823, opened 2015-11-05
Fix how joincluster interacts with ejabberdctl

joudinet:
Since joincluster needs access to both variables and functions from ejabberdctl, it is easier to source it within ejabberdctl, so we get access to all ejabberdctl's variables and functions inside joincluster.
As I was struggling with extra quoting required by EXEC_CMD, I decided to change it to a function that can prefix any command without adding any quote.
Hope it won't break any functionality I haven't tested.


cromain:
We're not using scripts anymore for clustering. So this PR is useless now.
see 15.11 blogpost

PR #849, opened 2015-11-25
mod_offline: Support discarding chat state notifications

### LINKED ISSUE #842: Chat state offline
Hi Guys,

I am working on an android app with smack. The problem is that chat states are stored offline. So if a user is offline and when he come online along with message he also get is typing notification. I can handle on my end by checking urn:xmpp:delay element but wanted to know if there is a better way.

Thanks,
Harshit

weiss:
XEP-0160 says that standalone chat state notifications should not be stored offline.  Let `mod_offline` support `store_empty_body: unless_chat_state`, which discards all chat states except for `<gone/>` notifications (as clients might [depend on those][1]).  Other messages with empty bodies (such as MUC invitations) are still stored offline unless `store_empty_body` is set to `false`.

Closes #842.

[1]: http://xmpp.org/extensions/xep-0085.html#bizrules-threads

### DIFF
--- a/src/jlib.erl
+++ b/src/jlib.erl
@@ -43,6 +43,7 @@
 	 get_iq_namespace/1, iq_query_info/1,
 	 iq_query_or_response_info/1, is_iq_request_type/1,
 	 iq_to_xml/1, parse_xdata_submit/1,
+	 is_standalone_chat_state/1,
 	 add_delay_info/3, add_delay_info/4,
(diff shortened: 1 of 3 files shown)


weiss:
I changed my mind and force-pushed a version of the commit that also discards `<gone/>` notifications.  This makes the behavior more straightforward, and my assumption is that droppping `<gone/>` notifications won't actually break things in practice.

PR #873, opened 2015-12-13
Update vcard_xupdate when clients sends presence stanza

sezuan:
For user who don't update their VCARD after activating this module an empty <photo/> tag will be added to the presence stanza. This will remove the avatar picture on certain clients like Gajim.

If the module receives a presence stanza and there's no entry in the vcards_xupdate table, it'll fetch the users photo and stores the hash in vcards_xupdate. If there's no PHOTO, the hash of the empty binary will be stored to avoid unecessary VCARD retrievals.


Neustradamus:
@sezuan: Why have you closed this PR?

PR #874, opened 2015-12-14
mod_irc: Disable spurious vCard->WHOIS requests.

kousu:
It makes sense that vCard in XMPP roughly maps to WHOIS in IRC,
but there is no code that actually handles receiving WHOISes properly:
they just get passed through as plain messages coming from user!ircserver@irc.xmppserver.tld.
I suppose sometimes you want to manually be able to run a WHOIS and see the results,
so I don't want them banned entirely, but I don't want them on each run.

Sending spurious WHOISes pisses off Xabber, which overenthusiastically sends "get vCard"
IQ stanzas on every single presence change (even "unavailable"); this has the horrible
side effect of spawning a huge number of private chats with users I have never talked to,
and buzzing my phone uncontrollably for a good minute, whenever I join a channel with more
than 50 people in it, and the same every time someone's connection drops.

Disabling this scratches an itch for me and patches over what is arguably Xabber's problem,
but in fact, this code was already half-disabled: _even though it sends IRC commands_ it
returns ?ERR_FEATURE_NOT_IMPLEMENTED.  Was this an oversight?

### DIFF
--- a/src/mod_irc_connection.erl
+++ b/src/mod_irc_connection.erl
@@ -451,8 +451,10 @@ handle_info({route_chan, Channel, Resource,
 					  ?ERR_FEATURE_NOT_IMPLEMENTED),
 	      ejabberd_router:route(To, From, Err);
 	  #iq{xmlns = ?NS_VCARD} ->
-	      Res = io_lib:format("WHOIS ~s \r\n", [Resource]),
-	      _ = (?SEND(Res)),
+	      % TODO: catch WHOIS replies and translate their details to something resembling vCards
+	      % For now this just errors out.
+	      %Res = io_lib:format("WHOIS ~s \r\n", [Resource]),
+	      %_ = (?SEND(Res)),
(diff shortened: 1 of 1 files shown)


kousu:
Huh. Why did that fail? The log just says
```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```
but it isn't the same error over and over, it's just very verbose. I don't understand how commenting out two lines could have made the build log verbose.

mremond:
@badlop I think `mod_irc` moved to contribs.
Still, maybe this patch can be useful.
What do you think ?

badlop:
Right, I'll commit this small change to the ejabberd-contrib/mod_irc repo.

badlop:
Committed three months ago in https://github.com/processone/ejabberd-contrib/commit/5c94639bf27d66bbc4ebcd2ec8f42729ec418087

PR #877, opened 2015-12-17
Adding support for WEBIRC and customize realname & ident.

iwalkalone69:
* Adding support for WEBIRC
* Adding support to customize realname & ident.
* Adding support to ISO-8859-15 and setting to default.


badlop:
Hi, I've fixed some indentation, recovered some DEBUG lines, and committed to ejabberd master Thanks!

PR #894, opened 2016-01-05
fix make release

genric:
add --enable-json configure param to include jiffy dependency
fix cache_tab name in reltool.config.script


cromain:
since 15.09, jiffy is fetched unconditionally (see 846b520) so we won't put --enable-json back.
but release needs fix indeed, i'll check all this.

cromain:
should be fixed by bbd428a (thanks to Pawel)

PR #900, opened 2016-01-12
Don't return error for blocked headlines and MUC messages

weiss:
If a message stanza is blocked as per XEP-0016 or XEP-0191, [return an error][1] only if the type of the blocked message is `normal` or `chat`. This makes sure users won't be kicked from MUC rooms when blocking other participants.

[1]: http://xmpp.org/extensions/xep-0016.html#protocol-error


mremond:
Good catch, I think this is a good idea.

PR #905, opened 2016-01-14
add room_id_regexp option

hamano:
I want to restrict room id such as ascii only.
This patch allow room_id_regexp option for muc.
room_id_regexp: "^[a-z0-9]+$"


badlop:
Thanks, I've applied your patch. But I also renamed the option to regexp_room_id, so take care to update your configuration file. It's also documented in the ejabberd Guide.

Issue #143, opened 2014-02-21
ejabberdctl.cfg ERL_OPTIONS make ejabberd crash


adelzhang:
for ejabberd tuning, add extra erlang options in ejabberd.cfg,

    ERL_OPTIONS="-mnesia dump_log_write_threshold 50000 -mnesia dc_dump_limit 40"

Then `ejabberdctl start` fail to start ejabberd, with CRASH REPORT

    2014-02-21 10:35:25 =ERROR REPORT====
    application_controller: syntax error before: '.': 
    2014-02-21 10:35:25 =CRASH REPORT====
      crasher:
        initial call: application_master:init/4
        pid: <0.40.0>
        registered_name: []
        exception exit: {{bad_return,{{ejabberd_app,start,[normal,[]]},{'EXIT',{aborted,{aborted,{application_load_error,{bad_environment_value,[]}}}}}}},[{application_master,init,4,[{file,"application_master.erl"},{line,133}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]}
    ancestors: [<0.39.0>]
    messages: [{'EXIT',<0.41.0>,normal}]
    links: [<0.39.0>,<0.7.0>]
    dictionary: []
    trap_exit: true
    status: running
    heap_size: 610
    stack_size: 27
    reductions: 112
      neighbours:

sawantuday:
Its still same with ejabberd 14.07.

developerworks:
`15.02` also has the issue.

developerworks:
`15.02` also has the issue.

weiss:
I guess we should [fix this][1] in `ejabberdctl`, but for now you should escape space characters with a backslash, e.g.:

    ERL_OPTIONS="-mnesia\ dump_log_write_threshold\ 50000\ -mnesia\ dc_dump_limit\ 40"

[1]: https://github.com/processone/ejabberd/pull/473

PR #907, opened 2016-01-14
don't escape ERL_OPTIONS

hamano:
in ejabberdctl.cfg:
ERL_OPTIONS="-pa /path/to/ebin" does not working.
no need to escape spaces.


badlop:
That line was introduced to solve https://github.com/processone/ejabberd/issues/143 And your commit does the opposite, right?

weiss:
I think the situation has changed since commit 98bad73d562f202c606f1aa8e51e1c2756cba15f, so this change is probably correct.

hamano:
I just understood that the issue has complicated history...
You can test the behaver:
$ ejabberdctl debug
(ejabberd@localhost)1> code:get_path().
Thank you.

PR #917, opened 2016-01-22
fix wrong timestamp

hamano:
ISO 8601 define basic format:
YYYYMMDDTHHmmss
and extended format:
YYYY-MM-DDTHH:mm:ss

but jlib:timestamp_to_iso() makes:
YYYYMMDDTHH:mm:ss

https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations
> Either basic or extended formats may be used, but both date and time must use the same format.

If you hate the change, It's a good idea that we make new functions timestamp_to_iso_basic() and timestamp_to_iso_extended().
Thank you.

### DIFF
--- a/src/jlib.erl
+++ b/src/jlib.erl
@@ -678,18 +678,18 @@ timestamp_to_iso({{Year, Month, Day},
                   {Hour, Minute, Second}},
                  Timezone) ->
     Timestamp_string =
-	lists:flatten(io_lib:format("~4..0B-~2..0B-~2..0BT~2..0B:~2..0B:~2..0B",
+	lists:flatten(io_lib:format("~4..0B-~2..0B-~2..0BT~2..0B~2..0B~2..0B",
 				    [Year, Month, Day, Hour, Minute, Second])),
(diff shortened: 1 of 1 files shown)


prefiks:
We have 1 and 2 argument version of this function and both do something different (2 argument version formats time stamp as defined in http://www.xmpp.org/extensions/xep-0082.html while version converts to older time stamp format as defined in http://xmpp.org/extensions/xep-0091.html). Switching name of 1 argument version to for example `timestamp_to_legacy_timestamp` would probably be not that bad idea, but  we still need both of those versions.

Additionally i think we can't remove those '-' and ':' from 2 argument version, per xep-0082 only variation that have those is correct.

So i can't accept this pull request, but if you would want to do that 1 argument version rename, i would gladly accept it.

hamano:
OK, I didn't see thir XEP.
I'll rename timestamp_to_iso/1 to timestamp_to_legacy/1 and
add new function timestamp_to_iso_basic/1.
Thank you!

PR #926, opened 2016-01-26
Remove empty and unused module

lpil:
Elixir modules with `.`s in the name are not nested, they're just module names with `.`s in them. As a result you don't need to define "parent" modules, unlike in languages such as Ruby.

Here I'm removing a module that is not needed.

### DIFF
--- a/lib/ejabberd.ex
+++ b/lib/ejabberd.ex
@@ -1,2 +1,0 @@
-defmodule Ejabberd do
-end


lpil:
Woo :)

mremond:
:+1:

lpil:
I think I'll put "ejabberd contributor" on my CV now ;P

mremond:
Removing lines of code is a science :)

PR #931, opened 2016-01-28
change mod_ping Timers using maps instead of dict

cclam0827:
Since OTP 18 maps function more efficient than dict, change the mod_ping using dict instead.

### DIFF
--- a/src/mod_ping.erl
+++ b/src/mod_ping.erl
@@ -44,8 +44,6 @@
 
 -define(DEFAULT_PING_INTERVAL, 60).
 
--define(DICT, dict).
-
 %% API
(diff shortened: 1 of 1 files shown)


weiss:
I'd use `maps:find/2` rather than `catch maps:get/2`.

cclam0827:
@weiss Thanks for your suggestion, I have changed to use `maps:find` instead :)

badlop:
Looking at the Erlang/OTP release notes, Maps was introduced in 17.0, and performance was improved in 18.0. ejabberd already requires 17.1, so this patch does not increase the requirements, and that's always a good thing :)

zinid:
@badlop we actually decided not to merge it

PR #943, opened 2016-02-03
Dependencies rework

mremond:
We now rely on new dependencies (fast_xml, fast_tls, etc) with more consistent naming.


lpil:
Would it be a good idea to draw the mix deps and the rebar deps from one file? Perhaps using `:file.consult/1` in the mix file to extract the deps from rebar?

mremond:
No, ejabberd full build is for now quite complex. Erlang build has many extra depedancies that are conditionnally build with configure flag. We have also a build chain for binary installers and packages based on this. I would not get into this at the moment. For now this should remains two separate builds for the sake of simplicity.

lpil:
Good to know, thank you.

PR #813, opened 2015-11-02
Support for oracle

softctrl:
I don't receive error using riak here, and also now i am using at least all modules (those with odbc support) with odbc, and also i make some tests using pgsql.

PR #979, opened 2016-03-03
A few changes to solve some problems with oracle connection ODBC.

softctrl:
A few changes to solve some problems with oracle connection ODBC.


mremond:
It seems the automatic tests does not pass with those changes.

softctrl:
hi, the problem appears because the original project use the name "timestamp" like a column name, we need to change this into the script. Many databases including Oracle do not allow us to use such name,  so to use this name on oracle i need to call with "timestamp", but this proccess solve the problem to Oracle, but create a error in mysql, because in the driver for oracle ODBC, the evaluate the correct for \"timestamp\", and send the value "timestamp" for the database. but in mysql appears that they send \"timestamp\" directly to the database, and then we have such error:

====
10:55:06.695 [error] SQL query 'insert into archive (username, "timestamp", peer, bare_peer, xml, txt, kind, nick) values ('test_master', '1458212106694768', 'test_slave@mysql.localhost/resource', 'test_slave@mysql.localhost', '<message from=''test_master@mysql.localhost/resource'' to=''test_slave@mysql.localhost/resource'' id=''14272383010905736991''><body>5</body></message>', '5', 'chat', '');' failed: "#42000You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '\"timestamp\", peer, bare_peer, xml, txt, kind, nick) values ('test_master', '1458' at line 1"
====

So, to solve the problem in the perfect way, please change all table that have reserved words. The best way to do this is simple insted use timestamp, we can concatenate with the table name like:

archive_timestamp


even use such method you need to be careful because all databases full ANSI/ISO SQL Standard  can not accept column names bigger than 30chars.


but this will solve all problems with databases incompatibility. 

Here some column names that cause troubles in databases full ANSI/ISO SQL Standard 

archive -> timestamp
rosterusers -> type
sm -> resource and priority

other problem, in a database full ANSI/ISO SQL Standard, the FROM clause is not an optional clause. So, the Microsoft SQL Server allow us to make such query:

"select 1;"

But, this is not compliant with ANSI/ISO SQL Standard. The FROM clause is always necessary, Because of that, oracle have a dummy table named DUAL. and we can do this:

"select 1 from DUAL;"

Here there is another problem.

softctrl:
so, of course i gladly can do this changes, but i think that such intrusive changes need to be made by some one that's know better all architecture from Ejabberd.

softctrl:
So, were the project is work great. i am sending some screen. I use Oracle because my client use this database.

image1

image2

softctrl:
so, the error

ERROR: No Elixir compiler found

apears to me like a problem on the machine.

bp91:
Hi @softctrl, Can I ask you the connection settings used for the oracle connection?
I tried with these:

```
sql_type: odbc
sql_server: "DSN=DB_NAME@//DB_HOST:DB_PORT/DB_SID;UID=ejabberd;PWD=ejabberd;"
```

```
sql_type: odbc
sql_server: "DSN=DB_NAME://DB_HOST:DB_PORT/DB_SID;UID=ejabberd;PWD=ejabberd;"
```
And I have this error:
```
[Driver Manager]DRIVER keyword syntax error SQLSTATE IS: IM012 Connection to database failed."
```
Can you help me?

Neustradamus:
@softctrl It is always possible for 18.12+?
It works?

Linked to:
- https://github.com/processone/ejabberd/pull/813
- https://github.com/processone/ejabberd/pull/795

Neustradamus:
@softctrl, have you looked for the master compatibility to add directly in the code?

Neustradamus:
@softctrl, have you looked for the master compatibility to add directly in the code?
@stealthgun would like...

Neustradamus:
@softctrl: Can you rebase and squash your PR?
@bp91, @stealthgun and me are very interested...

mremond:
Closing, as it would need some rework / rebase and testing.

Neustradamus:
@softctrl: Badly it has been closed.
Can you look to have a compatibility with current master?

Issue #1004, opened 2016-03-14
Incorrect 'from' attribute in roster push stanza  (mod_admin_extra)


oxoWrk:
version of ejabberd 16.01
debian 8.2
install from package

 I found same problem like #367 in mod_admin_extra ([push_roster_item](https://github.com/processone/ejabberd/blob/master/src/mod_admin_extra.erl#L1186))

PR #1011, opened 2016-03-15
Bare JID in 'from' of Roster Push (RFC 6121 section 2.1.6) in mod_adm…

oxoWrk:
fix for #1004


prefiks:
Looks good, will merge. Thanks for your contribution.

PR #1021, opened 2016-03-18
Use uuid for ejabberdctl

hamano:
This code work only linux but, 
There are several advantages: no concurrecy number limitation, no need flock, no need jot.
Also Free BSD has /compat/linux/proc/sys/kernel/random/uuid

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -387,7 +387,7 @@ ctl()
         # concurrent invocations using a bound
         # number of atoms
         for N in `seq 1 $MAXCONNID`; do
-            CTL_CONN="ejabberdctl-$N"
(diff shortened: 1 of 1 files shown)


cromain:
and osx comes with /usr/bin/uuidgen which produce same kind of output

cromain:
I'd better refactor this ctl function a bit so it just gets nodename from a dedicated function. If we can generalize counters and/or random (uuids) we can remove the flock/jot part to get a more generic and portable function.

cromain:
i've made some successful tests on a refactored (and simplified) ctl, using uniq ids. it works on osx/linux/netbsd/openbsd. similar approach is ok windows as well.
I may delay this to 16.04 if it don't have enough time for deeper tests.

PR #1024, opened 2016-03-18
Use the v0.2 release of luerl instead of a commit.

rbarlow:
The luerl project has made a tag for v0.2, which should be usable
by ejabberd. Discussion about the v0.2 release is here:

https://github.com/rvirding/luerl/issues/60


rbarlow:
I'm not sure if the failing test is due to my change or not. I'd appreciate someone taking a look and letting me know, because I wasn't expecting my commit to change the code in a meaningful way.

rbarlow:
It does appear that there are two commits that are not part of the v0.2 release that are part of the tag that was present before my change:

https://github.com/rvirding/luerl/commit/6ea78665fcdf8d5cca6da766156c3d685d28dbdb
and
https://github.com/rvirding/luerl/commit/9524d0309a88b7c62ae93da0b632b185de3ba9db

Perhaps this is a problem. FWIW, there is also a v0.3 release. Perhaps ejabberd can work with that instead?

prefiks:
It doesn't appear that your pull request caused this test failure, looks like for some reasons one test started to fail on erlang r17.

PR #1039, opened 2016-03-29
gen_mod stop modules for vhost

oxpa:
This is a simple patch which mimics gen_mod:start_modules(Host) and offers gen_mod:stop_modules(Host).


badlop:
I've implemented it in a slightly different way.

PR #1049, opened 2016-04-03
Fixing JOIN parse

iwalkalone69:
The parse of JOIN was wrong. Channel name its after :#, not before.


badlop:
Ah, right, the old code was something like:
```
> lists:subtract(":#cove", ":#").
"cove"
```
But the new code messed it when was converted to binary:
```
1>  [Chan1 | _] = binary:split(<<":#cove">>, <<":#">>), Chan1.
<<>>
```
However, your solution isn't yet fully equivalent:
```
2> [_ | Chan2] = binary:split(<<":#cove">>, <<":#">>), Chan2.
[<<"cove">>]
```
This would be better, right? Can you confirm this third solution is better?
```
3> [_, Chan3] = binary:split(<<":#cove">>, <<":#">>), Chan3.
<<"cove">>
```

iwalkalone69:
I am so sorry but I'm not good with erlang and I don't understand the difference with 2 and 3 in your examples. Now the code i submitted is (in bold the line changed:

    [FromUser | FromIdent] = str:tokens(From, <<"!">>),
    **[_ | Chan] = binary:split(Channel, <<":#">>),**
    ejabberd_router:route(jid:make(iolist_to_binary(

What would you change here?

Thanks.

badlop:
The difference between 2 and 3 is the character between _ and Chan: you put | and I put , because that way the result is more correct, apparently.

iwalkalone69:
You're right, what you say fixes JOIN and NICK messages.
Thanks a lot! I update the pull request.

mremond:
Hi, why did you close it ?

iwalkalone69:
Because I noticed it is failing and I want to submit it with more patches. In general I see that a lot of things have to be improved more closed to the IRC protocol. For example, KICK messages are not sending the information in a proper way for a web client.

mremond:
ok, fine then.

Please, keep us posted :)

iwalkalone69:
Maybe you can help me with some doubts. Why it is sending two stanza for some single IRC line? It could be done with only one. This happens in multiple places, for example process_kick().

mremond:
@badlop any idea about this question ?

iwalkalone69:
For now, I'm fixing some messages and adding some parameters to isolate parts of the IRC protocol messages. For example, in KICK, I've added the param moderator to be able to get it in the client as the one who kicks the user and message to get the kick message. What was happening also is that if kick message had :, this was truncated to the string since the last :.

I've also added message parameter to isolate the part and quit messages.

For now I'm planning to send the 366 numeric message (end of /names) to be able to, when client receives it, render all previous users & channel user count all at once in the web client jQuery app I'm coding to improve performance.

When I finish, I will post it here with all modifications and the reasons why I did it.

Issue #1053, opened 2016-04-06
Build fail on alpine - strinprep version


richp10:
Trying to compile 16.03 on alpine linux using Am using https://github.com/andyshinn/docker-alpine-abuild builder. 

Build failed with..

```
/home/builder/package/src/ejabberd-16.03/deps/stringprep/c_src/stringprep.cpp:38:12: warning: 'int compose(int, int)' defined but not used [-Wunused-function]
static int compose(int ch1, int ch2) {
^
ERROR: compile failed while processing /home/builder/package/src/ejabberd-16.03/deps/stringprep: rebar_abort
Makefile:90: recipe for target 'deps/.built' failed
```

Build worked perfectly if I upgraded  stringprep version in rebar.config from 1.02 to 1.03

cromain:
thanks for reporting this, i close this ticket to keep solution from #1054 and fix on 16.04

PR #1054, opened 2016-04-06
upgrade stringprep to 1.03

richp10:
Build fails on Alpine Linux with stringprep 1.02 - #1053

I have not run tests, but build works perfectly for me with the change to stringprep 1.03


mremond:
Thanks !

PR #1064, opened 2016-04-10
Fix check_password

sezuan:
check_password in mod_admin_extra doesn't provide authzid in ejabberd_auth:check_password call.


mremond:
It seems the tests were broken by the change :(

badlop:
Right, and I updated the tests immediately ;)

mremond:
@badlop very good, thanks !

PR #1074, opened 2016-04-18
fix broken check_password

sstrigler:
check_password when used from ejabberdctl/xmlrpc/rest etc is broken since 917d48f30bca65f984c4e1305eefe7266097ff65 was merged. This fixes it by adding ejabberd_auth:check_password/3 for backwards compatibility

### DIFF
--- a/src/ejabberd_auth.erl
+++ b/src/ejabberd_auth.erl
@@ -32,7 +32,7 @@
 -author('alexey@process-one.net').
 
 %% External exports
--export([start/0, set_password/3, check_password/4,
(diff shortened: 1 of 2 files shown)


beherit:
For me commit b67dc00db230fd2b0f1fda4345385b0e22567d7b already fixed this issue.

sstrigler:
If this is the prerequisite that this call exists in first place, also when looking at ejabberd_xmlrpc for instance, then you're right indeed.

sstrigler:
One side effect though: it breaks the deprecated mod_xmlrpc. If it's not already broken anyway.

That said you're free to close that PR here please.

PR #1080, opened 2016-04-21
specify su shell

hamano:
This script written in bash.
So it will fail depends login shell of INSTALLUSER.
Also we'll  get following error when setting /usr/sbin/nologin for login shell:
~~~
# ejabberdctl
This account is currently not available.
~~~

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -27,7 +27,7 @@ if [ "$INSTALLUSER" != "" ] ; then
                 mkdir -p "$INSTALLUSER_HOME"
                 chown "$INSTALLUSER" "$INSTALLUSER_HOME"
             fi
-            EXEC_CMD="su $INSTALLUSER -c"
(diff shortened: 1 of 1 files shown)


weiss:
You're using `su(1)`'s `--shell` flag which won't work on non-Linux systems.

hamano:
hmm, BSD su have -s option, but MacOS seems haven't it.
Also I tried to replace sudo but it does not work. cuz environ value does not passed.
~~~
            if hash sudo 2>/dev/null; then
                EXEC_CMD="sudo -u $INSTALLUSER bash -c"
            else
                EXEC_CMD="su $INSTALLUSER -c"
            fi

~~~
Ok, I'll just use `sudo ejabberdctl` so that I don't want to set login shell for system user.
You may want to ignore the trivial issue.
Thank you.

cromain:
indeed, i'll go the 'sudo -u ejabberd ejabberdctl' way as ejabberdctl will not be able to cope with all possible environments just using su

Issue #1078, opened 2016-04-21
ejabberd_c2s process didn't clear session data in some case


yzf:
> What version of ejabberd are you using?

16.03

> What operating system (version) are you using?

debian 8.0

> How did you install ejabberd (source, package, distribution)?

source

> What did not work as expected? Are there error messages in the log?

I use ejabberd for mobile app's im server, after running for a long time, I found that, the session number **keep increasing**, thought slowly. I debug into ejabberd's erlang shell, and found that **some session haven't been delete even thought the corresponding `ejabberd_c2s` process is terminate**. And there is **no error log**, it seems that those `ejabberd_c2s` terminate normally, except **leaving the session data behind**. Besides, the priority of all these sessions left behind are all `undefined`.

> was the unexpected behavior? What was the expected result?

The expected result is that `ejabberd_c2s` should delete the corresponding session data when it terminate.

---

After debug for a while, I found that there is some **logical error in the login progress**. 

Noticing that, in the commit https://github.com/processone/ejabberd/commit/a150bf8fdc8061bc60e5cb1a64a1e06f962d6a4e

the `session iq` is made to be optional, and session data is change to be created when receiving `iq bind`.

After receiving `bind iq` of the client, ejabberd create session data for it, and then send `iq result`. The problem is that, **if the connection is closed at that time, the send operation will fail**. Also,  `ejabberd_socket:send/2` will call `exit(normal)` if failed, this makes the `ejabberd_c2s` process jump into the `terminate/3` without update the `StateData`. So, **the state of this `ejabberd_c2s` process is still `wait_for_bind`, not `session_stablished`**. As a result, **it wouldn't delete its session data**.

zhchbin:
The steps to reproduce:
```diff
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1150,6 +1150,9 @@ wait_for_bind({xmlstreamelement, El}, StateData) ->
                                                       children =
                                                       [{xmlcdata,
                                                         jid:to_string(JID)}]}]}]},
+                                ?INFO_MSG("Sleeping....", []),
+                                timer:sleep(30000),
+                                ?INFO_MSG("Wake up....", []),
                                 send_element(StateData3, jlib:iq_to_xml(Res)),
                                 fsm_next_state_pack(
                                   session_established,
@@ -1918,6 +1921,7 @@ terminate(_Reason, StateName, StateData) ->
                 end,
                 bounce_messages();
             true ->
[...]

PR #1083, opened 2016-04-22
Fix issue #1078: ejabberd_c2s process didn't clear session data in some case

yzf:
Notice that, `send_element/2` will call `exit/1` if the connection is closed,
So, if create session data before response bind-iq-result,
the ejabberd_c2s process may jump into `terminate/3` without update state data.
As a result, it wouldn't delete the session data.

More details, see https://github.com/processone/ejabberd/issues/1078

In short, we should call `send_element/2` before `ejabberd_sm:open_session`.


zinid:
This has been fixed in a different way. Thanks for your contribution anyway!

PR #1087, opened 2016-04-24
Return error when blocking last activity request

weiss:
Return a generic service-unavailable error whenever an IQ-get or IQ-set is rejected (as per [RFC 6121][1], [XEP-0016][2], and [XEP-0191][3]).

[1]: https://tools.ietf.org/html/rfc6121#section-8.5.3.2.3
[2]: http://xmpp.org/extensions/xep-0016.html#protocol-error
[3]: http://xmpp.org/extensions/xep-0191.html#block

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -1563,6 +1563,12 @@ handle_info({route, From, To,
 							 {true, Attrs,
 							  StateData};
 						     deny ->
+							 Err =
+							     jlib:make_error_reply(Packet,
+										   ?ERR_SERVICE_UNAVAILABLE),
+							 ejabberd_router:route(To,
+									       From,
+									       Err),
 							 {false, Attrs,
 							  StateData}
 						   end;


zinid:
I'm not sure how to debug those service-unavailable received. We got a lot of questions like: "hey, I got service-unavailable, wtf?"

weiss:
Fair enough.  I've updated the PR so that it only adds a service-unavailable error for blocked last activity request (instead of dropping them silently) and leaves the other error messages alone.  The new error is again without an error text, as there's currently no text when blocking other IQ/message stanzas either.  Could add that if you like.

zinid:
We also probably can make sure any service-unavailable have error text except the ones related to privacy and presence. Presences will not be leaked in that case I think.

weiss:
> We also probably can make sure any service-unavailable have error text except the ones related to privacy and presence. Presences will not be leaked in that case I think.

Yes.  I think it already works that way, with the exception of groupchat messages (those [aren't easy to fix][1]) and IQ stanzas.  In my [original commit][2] I tried to avoid the leak for IQ stanzas by [removing an error text][3] from `ejabberd_sm`, but we _could_ instead wrap that one into a privacy list check and only include the error text if the traffic isn't blocked.

[1]: https://github.com/processone/ejabberd/commit/64d96778b452aad72349b21d2ac94e744617b07a
[2]: https://github.com/weiss/ejabberd/commit/f58c2eddaf0a5f12c7de8cb59809e7b2906b8a41
[3]: https://github.com/weiss/ejabberd/commit/f58c2eddaf0a5f12c7de8cb59809e7b2906b8a41#diff-f4ac2f1bb8ec3cea892d5378a71c045aL537

PR #1120, opened 2016-05-15
Send unique stanza id and archived tag also in the message carbons

anagromataf:
The archive UID of a message should also be in message carbons to allow the client receiving the carbons to identify the message correctly. Therefor the hooks of `mod_mam` should be invoked before `mod_carboncopy` to send the already archived message to the other clients.

- Change order of the hooks in mod_mam for sending and receiving packets. Messages are archived before a carbon copy is send to the other recourcces.
- Add archived tag and unique stanza id to the outgoing packet to have message carbons with the archive information.
- Add additional hook (in mod_mam) to strip the archive tag for outgoing packets after message carbons have been send.

### DIFF
--- a/src/mod_mam.erl
+++ b/src/mod_mam.erl
@@ -33,7 +33,7 @@
 %% API
 -export([start/2, stop/1]).
 
--export([user_send_packet/4, user_receive_packet/5,
+-export([user_send_packet/4, user_send_packet_strip_tag/4, user_receive_packet/5,
 	 process_iq_v0_2/3, process_iq_v0_3/3, disco_sm_features/5,
 	 remove_user/2, remove_room/3, mod_opt_type/1, muc_process_iq/4,
 	 muc_filter_message/5, message_is_archived/5, delete_old_messages/2,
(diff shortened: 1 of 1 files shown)


anagromataf:
Discussed this change already with @weiss. Maybe there is a more elegant solution to archive the same result.

anagromataf:
Does someone has any feedback on this?

weiss:
I think the fact that carbon copies aren't tagged with a stanza ID is clearly a bug that should be fixed. Having separate hook calls to add/strip the ID for outgoing carbons is of course not really nice though ...

weiss:
One alternative might be to move stanza ID creation/stripping from `mod_mam` into `ejabberd_c2s` (and `mod_muc_room`).  This would avoid the additional hook call, and the XEP-0359 tags might also be useful without MAM anyway.

cromain:
@anagromataf thanks for this PR. it looks ok for inclusion but i think we can check for possible improvements as suggested by @weiss for next release.
@weiss feel free to create a new ticket about improvements you have in mind.

PR #1132, opened 2016-05-25
Add support for PubSub publishing options

weiss:
Add code necessary to support publishing options as [described in XEP-0060][1].  A node plugin that expects publishing options must add `<<"publish-options">>` to the `features/0` list and then handle the publishing options handed over to the `publish_item/7` call.

This is one of the additions required to use the app server part of [`mod_push`][2].

[1]: http://xmpp.org/extensions/xep-0060.html#publisher-publish-options
[2]: https://github.com/royneary/mod_push

### DIFF
--- a/include/pubsub.hrl
+++ b/include/pubsub.hrl
@@ -93,7 +93,12 @@
 
 -type(subOptions() :: [mod_pubsub:subOption(),...]).
 
+-type(pubOption() ::
+    {Option::binary(),
+     Values::[binary()]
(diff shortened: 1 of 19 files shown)


zinid:
Wow, nice. I've been waiting for somebody to add this feature :)

zinid:
This is also will be needed by mod_mix.

cromain:
this patch is good, thanks Holger !

PR #1178, opened 2016-06-28
Harden the systemd unit

candrews:
Restrict capabilities, have a private tmp directory, private /dev, and don't accessing file system locations that really shouldn't be accessed.

This hardening improvement reduces the damage that can be done if ejabberd is compromised. There aren't any tests or anything like that that I can think of... to test it, just use the unit and make sure ejabberd continues to work :)

### DIFF
--- a/ejabberd.service.template
+++ b/ejabberd.service.template
@@ -12,6 +12,13 @@ ExecStop=@ctlscriptpath@/ejabberdctl stop
 ExecReload=@ctlscriptpath@/ejabberdctl reload_config
 Type=oneshot
 RemainAfterExit=yes
+# The CAP_DAC_OVERRIDE capability is required for pam authentication to work
+CapabilityBoundingSet=CAP_DAC_OVERRIDE
+PrivateTmp=true
+PrivateDevices=true
+ProtectHome=true
+ProtectSystem=full
+NoNewPrivileges=true
 
 [Install]
 WantedBy=multi-user.target


weiss:
# The CAP_DAC_OVERRIDE capability is required for pam authentication to work
    CapabilityBoundingSet=CAP_DAC_OVERRIDE

This actually gives ejabberd more privileges than it has with the current unit file.  ejabberd's PAM support installs a small `epam` binary to perform the actual authentication, so the setuid or capability privileges can and should be limited to this binary.

Apart from that, the changes look good to me.

candrews:
My understanding is that CapabilityBoundingSet expresses the maximum capabilities that can be acquired by this service in any way - it doesn't grant capabilities, it just allows them to be used if an attempt do so is done in a valid way. In this case, it doesn't mean that the binaries that run as the ejabberd user get this capability, but it does mean that if they run a suid binary (epam) then that can use the capability. (If CapabilityBoundingSet was set to empty, epam would get an error when I tried to do something requiring CAP_DAC_OVERRIDE, which is exactly what happened in my testing.) Therefore, this is definitely more secure than what the service does today.

weiss:
Ah sorry, you're completely right of course.  So all changes look good to me :-)

cromain:
Thanks ! this will be included in 16.08

weiss:
I've now seen multiple users stumble over `PrivateTmp=true` and `ProtectHome=true` when running commands such as `ejabberdctl restore /tmp/backup.dat`.

I'd suggest reverting both settings.  They are nice if you *know* that `/home` and `/tmp` won't be used to exchange data with the outside world, but that's not true for ejabberd.

candrews:
I'm unclear as to why people would be using /tmp in this way... Seems like bad practice that people should learn not to do. There is a real and good reason PrivateTmp exists - there have been many vulnerabilities reported for many years caused by shared /tmp.

The same idea applies to ProtectHome - logically, ejabberd has no business reading or writing to home directories. Sysadmin education is the way to address any problems people may be having - reducing security for everyone is not a good solution.

weiss:
> Sysadmin education is the way to address any problems people may be having [...].

I can imagine how such statements sound really good to those who aren't doing the actual ejabberd support (in their spare time).

> reducing security for everyone is not a good solution.

When dealing with real world software, you run into security vs. usablility trade-offs all day long.  And I think in this case, the potential security benefit of these options isn't worth letting many (*many*!) admins stumble over totally unexpected behavior.

You may opt one way or another.  But this *is* a trade-off.  So please don't pretend it's a "right or wrong" thing where the choice would be obvious.

debalance:
@candrews, what's your take on https://bugs.debian.org/836711 ?

candrews:
I'm not using Debian and I can't reproduce the problem. I am using ejabberd with this systemd unit and pam and it works fine (I'm on Gentoo).

`CapabilityBoundingSet=CAP_DAC_OVERRIDE` does not broaden privileges. CapabilityBoundingSet determines the bounding set (in other words, maximum privileges) - it does not actually assign any privileges.

Another thing to considering is reaching out to the Debian systemd team - see the very bottom of https://wiki.debian.org/Teams/pkg-systemd/Packaging for that contact info. My apologies, but I'm not sure how pam is set up on Debian so I am not sure how to help in this situation - but I'm sure the Debian systemd people have encountered similar situations and can provide assistance.

Issue #625, opened 2015-06-29
Writing Elixir modules


dapriett:
Hello,

I'm new to both writing modules for Ejabberd as well as Elixir (and erlang).  I'm trying to follow the blog post here: https://blog.process-one.net/elixir-sips-ejabberd-with-elixir-part-1/

I'm using the Mac OSX installer of ejabberd (https://www.process-one.net/en/ejabberd/downloads/) - how would I add a custom Elixir module into it?  Do I just create a .ex Elixir module and dump it into the lib directory?  Is there some sort of compilation I have to do first?  Do you have a sample Elixir module project out there?

Thanks, and sorry for the newbie questions.

mremond:
Well, if you want to develop on Elixir with ejabberd, basically, the blog post shows that you need a development environment, source code, etc.
We never tried yet developing Elixir module from the binary installer. It is intended to be a small, ready to use package for users, not for developers. Probably you will be missing development tool.

My recommendation is thus to install Erlang, Elixir on your machine and to build ejabberd from source. At the moment, that's the best approach.

I will keep the ticket open to see if we can test the build chain for Elixir and see if we can make it work.

dapriett:
Ok thanks. 

I tried copying my elixir project to my ~/.ejabberd-modules/sources.  I had to create a .spec file to make it see it, and I was able to do module_install but it looks like the install creates an empty ebin directory.  So guessing the build chain doesn't support it yet.

mremond:
Yes the module installer does not yet support Elixir module.

nrako:
I'd love to see support of Elixir module by the module installer. Or else some guide to compile a Elixir module for ejabberd independently `elixirc -o ebin -I ....` without having to rely on the main ejabberd build would also help people like me who are new to this env.

KadekM:
+1 on support from module installer :)

sebyx07:
also a better documentation on behaviours and other basic stuff you can do from erlang/elixir would be great

cromain:
Feature added by #1193 thanks to Gabriel

PR #1193, opened 2016-07-06
Fix issue #625: Writing Elixir modules

gabrielgatu:
PR for supporting elixir modules in the installer. Resolves issue #625.

It compiles all the `*.ex` files found inside `/lib` and `/deps` folder.

I've tried to leverage the `Kernel. ParallelCompiler ` module from elixir instead of executing a cmd command (using elixirc), but to maintain consistency with the error checking I'm passing one file at a time  instead of passing them all at once, not exploiting the parallel compilation of elixir. This has been necessary because the `files_to_path/3` function doesn't return anything when a compilation error occurs, instead it throws an exception.

### DIFF
--- a/src/ejabberd_app.erl
+++ b/src/ejabberd_app.erl
@@ -45,6 +45,7 @@ start(normal, _Args) ->
     write_pid_file(),
     jid:start(),
     start_apps(),
+    start_elixir_application(),
     ejabberd:check_app(ejabberd),
     randoms:start(),
     db_init(),
@@ -237,3 +238,9 @@ opt_type(modules) ->
 		      Mods)
     end;
 opt_type(_) -> [cluster_nodes, loglevel, modules, net_ticktime].
+
+start_elixir_application() ->
+  case application:ensure_started(elixir) of
+    ok -> ok;
+    {error, Msg} -> ?ERROR_MSG("Elixir application not started.", [])
+  end.
--- a/src/ext_mod.erl
+++ b/src/ext_mod.erl
@@ -484,17 +484,28 @@ compile_deps(_Module, _Spec, DestDir) ->
     filelib:ensure_dir(filename:join(Ebin, ".")),
     Result = lists:foldl(fun(Dep, Acc) ->
                 Inc = filename:join(Dep, "include"),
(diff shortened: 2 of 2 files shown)


mremond:
Do you have an example Elixir module we could had to ejabberd-contrib repository to test this patch ?
That would be helpful.
Thanks !

gabrielgatu:
Just create a folder inside /ejabberd-contrib with a:
  - ./mod_name.spec
  - ./lib/elixir_modules....

The only difference with erlang modules is that is expecting the files to be located inside the /lib folder instead of /src. You can also have both erlang and elixir files, inside their respective folders.

This is a sample module you can test. Just put it inside ~/.ejabberd-modules/sources/ejabberd-contrib.

[mod_elixir.zip](https://github.com/processone/ejabberd/files/377805/mod_elixir.zip)

mremond:
We need the to start Elixir before. We can add call to start Elixir application from other branch (warning typo on `_application`):
https://github.com/processone/ejabberd/pull/1208/files#diff-edea3bc6cafa88f1fd6385737dc59ddcR58

and the actual method: https://github.com/processone/ejabberd/pull/1208/files#diff-edea3bc6cafa88f1fd6385737dc59ddcR256

If you could add this small things to the PR branch, Christophe thinks we are ready to merge.

Thanks !

cromain:
indeed, just adding ejabberd:start_app(elixir). in ejabberd_app:start_apps/0 should be enough

cromain:
hum i was a bit optimistic with simple approach of start_app. if elixir is not compiled then ejabberd won't start with such simple patch.
replacing ejabberd:start_app by application:start will fix this (we do not control result of operation and continue even if elixir is not found)

gabrielgatu:
Hi Christophe !

first of all, nice to meet you :) I’ve talked with mickael and he agreed to integrate a solution I’ve worked on regarding this. I’m preparing the commit right now :)
I would like to hear what do you think about it!

Gabriel.

cromain:
Hi Gabriel !

nice to see your improvements around elixir modules :)
this last patch looks perfect. I think it's OK for merge now

PR #1201, opened 2016-07-18
External Component Connection Hooks

xmppjingle:
External Component Connection Hooks 

Event List:
component_connected(ComponentAddress) -> ok
component_disconnected(ComponentAddress, Reason) -> ok


zinid:
Isn't it better to call `component_disconnect` hook in `terminate/3` when StateName is `session_established` (along with de-registering route)? Because in your version if an `ejabberd_service` process fails the hook won't be called.

xmppjingle:
Thanks for the suggestion. I will make the modifications and push again.

xmppjingle:
Suggestion applied and Reason added upon disconnection

mremond:
Thanks !

PR #1208, opened 2016-07-22
Support for Elixir configuration file

mremond:
This PR is intended to track changes made for GSOC 2016 project.


config/ejabberd.exs
```diff
@@ -0,0 +1,169 @@
+defmodule Ejabberd.ConfigFile do
```
mremond: I would add here:

```
import Ejabberd.Logger # this allow using info, error, etc for logging
```

Probably nice to add to the template, or even auto-import when you use `Ejabberd.Config`.

lib/ejabberd/config/ejabberd_module.ex
```diff
@@ -0,0 +1,67 @@
...
+  It offers functions for validation and for starting the modules.
+
+  Warning: The name is EjabberdModule to not collide with
+  the already existing Elixir.Module.
+  """
+
+  @type t :: %{module: atom, attrs: [Attr.t]}
+
+  defstruct [:module, :attrs]
+  alias Ejabberd.Config.{EjabberdModule, Attr, Validation}
```
mremond: This line does not compile with Elixir 1.1.1.
https://travis-ci.org/processone/ejabberd/jobs/146668610#L1700

Is it possible to use another syntax to make the code portable across more version of Elixir ?

src/ejabberd_app.erl
```diff
@@ -55,6 +55,8 @@ start(normal, _Args) ->
     ejabberd_admin:start(),
     gen_mod:start(),
     ext_mod:start(),
+    start_elixir_applicaton(),
```
mremond: Typo: start_elixir_applicaton -> start_elixir_application

mremond:
This is merged in: https://github.com/processone/ejabberd/commit/803270fc6b8ed3ba718f7e231b149caef70aa1ae

Thanks !

PR #1211, opened 2016-07-23
Expand api

mremond:
For review / comments. No ready to merge yet.

This PR improves error reporting for ejabberd OAuth API and allow checking conformance of commands against an ACL.


prefiks:
I personally would prefer if we would have access rules specified in config file, instead of hardcoding that in sources (even we sometimes need different for saas and public version). Right now only have one list of command with one access rule, but i think we could implement something like that:

```
commands_access:
  - group:
    - commands:
      - register
      - status
    - access:
      - ip: "127.0.0.1/8"
  - group:
    - commands:
       - stop
    - access: admins
```

If we later add acl based on oauth scope i think this will make good match to that.

prefiks:
But parts that improve error reporting i like.

PR #1216, opened 2016-07-26
Add functions to remove whitespace

xmppjingle:


mremond:
It now seem the scope of the PR include other stuff, like packaging. Maybe it is best to rework them to have more unique scope ?

xmppjingle:
My bad, switched branch and end up pushing to wrong one. Will fix this asap.

PR #1225, opened 2016-08-01
Whitespace remover

xmppjingle:
Whitespace Remover Function added to lib module


zinid:
As `jlib.erl` is deprecated now, this PR now is incompatible.
Please create the PR against new version if needed.

PR #1233, opened 2016-08-08
Set MemoryDenyWriteExecute in the systemd unit

candrews:
I've tested this changed and it seems to work great. Note that because systemd ignored unknown options, this change will not cause problems for users of systemd < 231.

The benefit is that, for users of systemd 231, a whole class of attacks are mitigated.

### DIFF
--- a/ejabberd.service.template
+++ b/ejabberd.service.template
@@ -19,6 +19,7 @@ PrivateDevices=true
 ProtectHome=true
 ProtectSystem=full
 NoNewPrivileges=true
+MemoryDenyWriteExecute=true
 
 [Install]
 WantedBy=multi-user.target


weiss:
The `MemoryDenyWriteExecute` [documentation][1] says that

> a system call filter is added that rejects `mmap(2)` system calls with both `PROT_EXEC` and `PROT_WRITE` set, `mprotect(2)` system calls with `PROT_EXEC` set and `shmat(2)` system calls with `SHM_EXEC` set.  Note that this option is incompatible with programs that generate program code dynamically at runtime, such as JIT execution engines, or programs compiled making use of the code "trampoline" feature of various C compilers.

I would assume this will at least break [HiPE][2]-compiled ejabberd instances, and [maybe][3] also non-HiPE builds.

[1]: https://www.freedesktop.org/software/systemd/man/systemd.exec.html
[2]: https://www.it.uu.se/research/group/hipe/
[3]: https://github.com/erlang/otp/blob/master/erts/emulator/sys/common/erl_mmap.c#L1350

cromain:
This is job of distribution packagers to improve/tune our default systemd unit depending of distribution choices. We won't merge this change in mainline.

PR #1247, opened 2016-08-15
Add redis support for mam

sngyai:


mremond:
Does it makes sense to store MAM messages in Redis ? Redis is an in-memory db and MAM are high volume data. Why do you need it ?

PR #1253, opened 2016-08-17
Xep0356

Amuhar:
This PR is for  "Privileged Entity Support to Write Powerful External Components for ejabberd" project of gsoc 2016.

It was tested using sat components. 

It is documented here https://github.com/Amuhar/docs-for-xep0355-and-0356


cromain:
I merged this PR as of af0a493 to cleanup few indendation/format changes unrelated to the feature.

zinid:
I would personally revert this PR.
Here are the problems with it:
* Direct calls to mod_* modules in several places, for example [here](https://github.com/processone/ejabberd/blob/master/src/ejabberd_service.erl#L251), [here](https://github.com/processone/ejabberd/blob/master/src/ejabberd_service.erl#L252) and so on. Totally 13 (!!!) places. Those should be done using hooks.
* 'Meta' hook registration, for example [here](https://github.com/processone/ejabberd/blob/master/src/ejabberd_service.erl#L553). This will not allow automated software to detect incorrect hooks registration. I have already eliminated all such places in ejabberd, yet this PR is inroducing new ones.
* Hook registrations on anonymous functions (see previous point). While this is allowed by ejabberd, the main drawback is that anonymous functions being permanently stored won't make it possible to reload code.
* Using `ejabberd_service:send_element/2` from foreign modules, e.g. [here](https://github.com/processone/ejabberd/blob/master/src/mod_privilege.erl#L48). This is a bad idea, because 'sending' API is a subject for frequent changes. I'm failing to see why `ejabberd_router:route/3` couldn't be used in this situation.
* Sending to bare pids, e.g. [here](https://github.com/processone/ejabberd/blob/master/src/mod_privilege.erl#L71). Read [this](http://www.erlang.se/doc/programming_rules.shtml#HDR21) for the explanation why this is a bad idea.
* `ejabberd_service` now knows too much and does too much job. For example, it now registers hooks for foreign modules and keeps state for foreign modules. While keeping additional state is ok, this should be done on behalf of foreign modules, e.g. `ejabberd_service` should not set/get foreign values in its state.
* No tests are written, despite we have common_test framework for ages now. This is now impossible to refactor this code.

Fixing all these issues requires some time and effort. This should be done by the PR's author, we don't have to take this burden ourselves.

PR #1254, opened 2016-08-17
Remove no longer necessary p1_xmlrpc

lemenkov:
Ths module was superseded by fast_xml in commit
processone/ejabberd@dfc29ea03ca91e1eb5387d93612e2ac4b4b496da. So let's remove it entirely to
avoid any confusion.

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>


lemenkov:
Note - these failures does not looks like related to this patch. They are Elixir-related.

PR #1261, opened 2016-08-22
Fixed #1260 Stream Management feature for the websocket connections

### LINKED ISSUE #1260: Can't disable Stream Management feature for the websocket connections

16.03

Debian Jessie

source

I did try to disable Stream Management (XEP-0198). To achieve that, I specified false for the stream_management option in the config file.
But it doesn't make any effect when a client opens XMPP session via websockets. 
When a client (Gajm, Pidgin) establishes connection directly, it works like expected.

iAchilles:
Settings to disable Stream Management feature for the websockets connections:

```
port: 5280
    ip: "::"
    module: ejabberd_http
    request_handlers:
      "/ws-chat": ejabberd_http_ws
    stream_management: false
```


weiss:
Thanks!

Issue #1271, opened 2016-08-25
Resource gets stuck sometimes after client disconnect


sabudaye:
> What version of ejabberd are you using?

16.08

> What operating system (version) are you using?

`
-bash-4.1$ cat /etc/redhat-release
CentOS release 6.7 (Final)
`
`
-bash-4.1$ uname -a
Linux 2.6.32-573.18.1.el6.x86_64 #1 SMP Tue Feb 9 22:46:17 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
`

> How did you install ejabberd (source, package, distribution)?

source

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

Client opens connection with stream management without resumption. After client closes connection (in any ways), sometimes server has online resource for this user (despite resume_timeout for c2s is set to 0) without any reasonable timeout, so that means offline messages never be stored to "spool" table, and never be sended to user after his reconnection.

ejabberd.yml:

```
loglevel: 5

log_rotate_size: 10485760
log_rotate_date: ""
log_rotate_count: 1
log_rate_limit: 100

hosts:
  - "some.host"

listen:
  -
    port: 5222
    module: ejabberd_c2s
    resend_on_timeout: if_offline
    max_ack_queue: 3000
    max_fsm_queue: 3000
    max_resume_timeout: 30
    resume_timout: 0
    max_stanza_size: 65536
    shaper: c2s_shaper
    access: c2s
  -
    port: 5269
    module: ejabberd_s2s_in

  -
    port: 5280
    module: ejabberd_http
    request_handlers:
      "/websocket": ejabberd_http_ws
    web_admin: true
    http_bind: true
    resend_on_timeout: true
    captcha: true

auth_method: external
extauth_program: "/some/path/to/external_auth.sh"
extauth_instances: 4

odbc_type: pgsql
odbc_server: "some.pg.host"
odbc_database: "some_ejabberd_db"
odbc_username: "some_user"
odbc_password: "some_password"
odbc_port: 6432

redis_server: "some.redis.host"
redis_port: 63003
redis_db: 5
sm_db_type: redis

shaper:
  normal: 3000
  fast: 50000

max_fsm_queue: 3000

acl:
  admin:
    user:
      - "admin": "localhost"
  local:
    user_regexp: ""
  loopback:
    ip:
      - "127.0.0.0/8"

access:
  max_user_sessions:
    all: 50
  max_user_offline_messages:
    admin: 5000
[...]

PR #1282, opened 2016-09-05
fix is_online/1 function to filter sessions without living process

sabudaye:
in situation which I described in #1271 I can see this output in debug console:

```
(ejabberd@localhost)1> ejabberd_sm_redis:get_sessions(<<"user_1136">>, <<"some.host">>, <<"ru.app.debug_1.56.0-D_build_2_Android_5.0_62df618a-cc30-4d24-b3bf-216fcf810126.1470982580701">>).
[{session,{{1473,54402,3278},<0.8407.4>},
          {<<"user_1136">>,<<"some.host">>,
           <<"ru.app.debug_1.56.0-D_build_2_Android_5.0_62df618a-cc30-4d24-b3bf-216fcf810126."...>>},
          {<<"user_1136">>,<<"some.host">>},
          0,
          [{ip,{{85,140,6,7},31920}},
           {conn,c2s},
           {auth_module,ejabberd_auth_external}]}]
```

meanwhile:

```
(ejabberd@localhost)2> is_process_alive(list_to_pid("<0.8407.4>")).
false
```

To fix this issue, I think we can add additional check to is_online/1 function, to be sure that session process is still alive.

### DIFF
--- a/src/ejabberd_sm.erl
+++ b/src/ejabberd_sm.erl
@@ -432,8 +432,14 @@ online(Sessions) ->
 
 -spec is_online(#session{}) -> boolean().
 
-is_online(#session{info = Info}) ->
-    not proplists:get_bool(offline, Info).
+is_online(#session{info = Info, sid = Sid}) ->
+    not proplists:get_bool(offline, Info) andalso is_session_alive(Sid).
+
+-spec is_session_alive(sid()) -> boolean().
+
+is_session_alive(Sid) ->
+	Pid = element(2, Sid),
+	rpc:call(node(Pid), erlang, is_process_alive, [Pid]).
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 


weiss:
> To fix this issue, I think we can add additional check to is_online/1 function, to be sure that session process is still alive.

That sounds like hiding a bug to me.  The question is why the session wasn't removed from Redis.

sabudaye:
[documentation](http://erlang.org/doc/man/gen_server.html#Module:terminate-2) says:
```
If the gen_server process is part of a supervision tree and is ordered by its supervisor to terminate, this function is called with Reason=shutdown if the following conditions apply:

    The gen_server process has been set to trap exit signals.

    The shutdown strategy as defined in the child specification of the supervisor is an integer time-out value, not brutal_kill.

...

Otherwise, the gen_server process terminates immediately.
```
So that means, sometimes terminate callback does not called, and its normal behavior. Anyway, I think we can not blindly trust to data returned from session storage backend.

weiss:
> So that means, sometimes terminate callback does not called, and it is normal behavior.

An `ejabberd_c2s` process receiving an exit signal is not "normal behavior".

And I haven't ever seen dangling session table entries on any of my servers, nor have I seen others report this issue.  So you're running into some unknown issue, and your change is hiding it.

cromain:
I think we should close this PR
as @weiss said, it just hide the root cause of the issue. it's not a fix.

PR #1287, opened 2016-09-07
New stream management option: ack_timeout

weiss:
Close the connection if a stream management client fails to respond to an acknowledgement request within 60 seconds.  This number of seconds can be changed with the new `ack_timeout` option, and the mechanism can be disabled by specifying `infinity`.

As a side effect of this change, a new acknowledgement is no longer requested before the response to the previous request is received.

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -113,9 +113,12 @@
 		mgmt_pending_since,
 		mgmt_timeout,
 		mgmt_max_timeout,
+		mgmt_ack_timeout,
+		mgmt_ack_timer,
 		mgmt_resend,
 		mgmt_stanzas_in = 0,
 		mgmt_stanzas_out = 0,
+		mgmt_stanzas_req = 0,
 		ask_offline = true,
 		lang = <<"">>}).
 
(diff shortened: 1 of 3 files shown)


src/ejabberd_c2s.erl
```diff
@@ -1759,6 +1768,11 @@ handle_info({broadcast, Type, From, Packet}, StateName, StateData) ->
     fsm_next_state(StateName, StateData);
 handle_info(dont_ask_offline, StateName, StateData) ->
     fsm_next_state(StateName, StateData#state{ask_offline = false});
+handle_info(close, StateName, StateData) ->
+    ?DEBUG("Timeout waiting for stream management acknowledgement of ~s",
+	   [jid:to_string(StateData#state.jid)]),
+    close(self()),
+    fsm_next_state(StateName, StateData);
```
prefiks: Shouldn't you set mgmt_ack_timer to undefined here, won't it be still set in case that session is resumed later?

weiss: I don't think that's necessary, as the connection that resumes the session has a new fresh `#state` and `mgmt_ack_timer` is not [inherited from the old state][1].  But we could of course set `mgmt_ack_timer: undefined` to clarify the code.

[1]: https://github.com/weiss/ejabberd/blob/621f0e2b7cbb6c431e2c26ad6561c6c0eb74c065/src/ejabberd_c2s.erl#L3097

weiss: > But we could of course set `mgmt_ack_timer = undefined` to clarify the code.

I was going to update the PR accordingly, but there's the minor downside that a new timer would be spawned if another stanza arrives before the process actually goes into the `wait_for_resume` state.  This is avoided by *not* setting `mgmt_ack_timer = undefined`.  We could set it to `expired` or something, though :-)

prefiks: That's fine, i just was not sure how new session is created from old one, if this will not affect it, then i am ok with it.

prefiks:
Looks ok to me, so let's merge it.

Issue #1316, opened 2016-10-03
Option for disabling postgresql prepared statements


sabudaye:
> What version of ejabberd are you using?

16.08

> What operating system (version) are you using?

`
-bash-4.1$ cat /etc/redhat-release
CentOS release 6.7 (Final)
`
`
-bash-4.1$ uname -a
Linux 2.6.32-573.18.1.el6.x86_64 #1 SMP Tue Feb 9 22:46:17 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
`

> How did you install ejabberd (source, package, distribution)?

source

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

We use PgBouncer in transaction pooling mode, and only option to deal with prepared statements is disabling them. It would be great to have configuration option in ejabberd.yml for that.

alexeyshch:
Thank you, your changes are now in master.

PR #1317, opened 2016-10-03
Added option to disable prepared statements for postgresql

sabudaye:
#1316


weiss:
I didn't look into the actual change, just some random comments:

- According to Travis, the build currently fails with `src/ejabberd_sql.erl:526: variable 'Host' is unbound`.
- The `ejabberd.yml.example` diff contains many lines where only whitespace is changed.
- I'd rather name the option `sql_prepared_statements` and document that it currently only affects PostgreSQL.

sabudaye:
@weiss fixed. Not sure about ?MYNAME macros, but it was already used in ejabberd_auth_* modules in the same case

PR #1326, opened 2016-10-07
Enable elixir when used as a mix dependency

ukutaht:
When using ejabberd as a mix dependency, pass the appropriate compiler flag to make

``` erlang
-ifdef(ELIXIR_ENABLED).
```

preprocessor macros true.

This allows users to write config files in elixir when ejabberd is embedded in a mix project. @gabrielgatu I see you've done most of the work on elixir config. Does this look right?


gabrielgatu:
Seems good to me, but I really don't have the necessary experience with make to evaluate properly the impact of this change, so I'll leave the decision to someone more experienced than me :)

cromain:
this is now applied, thanks !

PR #1337, opened 2016-10-13
Don't send the whole MUC backlog by default

l29ah:
…when a client doesn't specify the amount of messages it wants, so it
becomes possible to have huge backlogs, so `since`-aware clients are
able to fetch outstanding messages.

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -2207,7 +2207,11 @@ count_stanza_shift(Nick, Els, StateData) ->
 	       false -> 0;
 	       _ -> count_maxchars_shift(Nick, MaxChars, HL)
 	     end,
-    lists:max([Shift0, Shift1, Shift2, Shift3]).
+    ShiftST = lists:max([Shift0, Shift1, Shift2, Shift3]),
+    case ShiftST of
+	       0 -> count_maxstanzas_shift(40, HL);
+	       _ -> ShiftST
+	     end.
 
 count_seconds_shift(Seconds, HistoryList) ->
     lists:sum(lists:map(fun ({_Nick, _Packet, _HaveSubject,


badlop:
In this patch you are forcing a maxstanzas of 40. Why 40? And why forcing it? There's already an option for this, called history_size. And its default value is 20, so this patch isn't required by default.

If the sysadmin sets a big high value, then obviously he is provoking a problem in the clients. In such case, clients can use maxstanzas http://xmpp.org/extensions/xep-0045.html#enter-managehistory

So, I suspect you are abusing the discussion history feature, and you are breaking your clients. Discussion history is designed to be only the most recent history, and clients may limit it even further. It isn't a room full logging, with short defaults that clients can increase... Maybe you want to use mod_mam instead.

zinid:
I didn't understand the meaning of the patch as well :/
From what I have understood: if we receive empty `<history/>` element, then we send last 40 messages of history. Makes no sense to me.

l29ah:
> It isn't a room full logging, with short defaults that clients can increase...
Why? It's simple and convenient, and XEP allows for it.

> From what I have understood: if we receive a <history/> element which doesn't match any history elements, then we send last 40 messages of history. Makes no sense to me.
If the client doesn't ask any concrete amount of history elements, ejabberd sends the whole history by default that can be troublesome bandwidth-wise and generally unexpected. This patch limits it to 40 entries. Maybe i should've used '20' to align with the ejabberd's default MUC history size.

badlop:
I understand your usercase is an admin that sets history_size to 1000 or 5000, and then this limit to 40 (or 20). This way, most clients receive just the 40-lines recent history, and some clever high-bandwidth clients can request the 1000-lines full history.

But notice that this patch modifies the history-size option: it no longer defines the number of messages received by the clients, because you are forcing the return to 40 (or 20, or whatever the code says).

As I said, if someone wants clever high-bandwidth clients to receive a large room history (much more than the human reasonable), he can enable mod_mam and let the clients request exactly the timeframe they want.

l29ah:
On Mon, Oct 17, 2016 at 08:37:11AM -0700, badlop wrote:
> I understand your usercase is an admin that sets history_size to 1000 or 5000, and then this limit to 40 (or 20). This way, most clients receive just the 40-lines recent history, and some clever high-bandwidth clients can request the 1000-lines full history.
> 
> But notice that this patch modifies the history-size option: it no longer defines the number of messages received by the clients, because you are forcing the return to 40 (or 20, or whatever the code says).
> 
> As I said, if someone wants clever high-bandwidth clients to receive a large room history (much more than the human reasonable), he can enable mod_mam and let the clients request exactly the timeframe they want.

Okay, i guess i should make another option for that.

-- 
()  ascii ribbon campaign - against html mail
/\  http://arc.pasp.de/   - against proprietary attachments

PR #1346, opened 2016-10-19
Improve systemd unit file

### LINKED ISSUE #1281: pam auth fails after upgrading from 16.06 to 16.08

16.08

Debian jessie 8.5

jessie-backports package version 16.08-1~bpo8+1

pam authentication does not work, the epam process is running as ejabberd:ejabberd even tho the binary is setgid shadow. Without group shadow pam cannot access /etc/shadow and so it can't authenticate the user.

I've had a dig around in the ejabberd and epam source code and can't see anything obvious unfortunately.

Note that the epam binary was not changed by the upgrade (or downgrade to 16.06)

Debian bug report is here:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=836711

### LINKED ISSUE #1297: mod_http_upload docroot failed

16.08

Debina 8.5 (OpenVZ container)

deb https://apt.jabber.at jessie ejabberd                                                                                                                                             

Сeased to upload files after update

```
2016-09-14 14:19:07.060 [error] <0.637.0>@mod_http_upload:process:428 Cannot store file /home/ejabber/storage/***/***/***.jpg from ::FFFF:***.***.*.* for upload.thfree.ru: "permission denied"
```

If you change the root directory

/etc/ejabberd/ejabberd.yml

```
  mod_http_upload:
...
#    docroot: "/home/ejabber/storage/"
    docroot: "/tmp/"
...

```

it creates a

```
/tmp/systemd-private-d09f46d7664643bd81ded58a2c506075-ejabberd.service-fkQ25u/tmp/***/***/***.jpg
```

### LINKED ISSUE #1347: Problem with PID file and setcap

16.09

Debian 8.6

From package provided by jabber.at.

After upgrading from version 16.04 to 16.09 ejabberd stops writing PID file - I spent a few hour to fix it and I found that the problem lies with caps that I set for erlang:

```
setcap 'cap_net_bind_service=+ep' /usr/lib/erlang/erts*/bin/beam
setcap 'cap_net_bind_service=+ep' /usr/lib/erlang/erts*/bin/beam.smp
```

When I removed caps ejabberd start writing PID file again. I don't upgraded erlang with upgrading ejabberd, is there any way to fix it?

P.S. I need this to set listening port 80 and 443 in ejabberd.

### LINKED ISSUE #1376: NoNewPrivileges=true disallows ejabberd to start on systems with SELinux enforcing

ejabberd-16.08-2.fc25.x86_64

Fedora 25 (pre-release)

distribution

Fedora has been using a systemd unit file very similar to the [ejabberd.service.template](https://github.com/processone/ejabberd/blob/16.09/ejabberd.service.template) file provided in this repository. There is an [issue filed against systemd](https://github.com/systemd/systemd/issues/3845) since this wasn't clear in its documentation that reports that ```NoNewPrivileges=true``` will lead to denials upon the process attempting to alter its context:

```
type=SELINUX_ERR msg=audit(1478750215.370:619): op=security_bounded_transition seresult=denied oldcontext=system_u:system_r:init_t:s0 newcontext=system_u:system_r:rabbitmq_t:s0
```

For now, I recommend dropping the ```NoNewPrivileges=true``` line from the unit file. I decided to file an issue instead of a pull request in case discussion was warranted first.

This downstream issue tracks this problem in Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=1393587

weiss:
Apply a number of improvements to the systemd unit file:
- Don't let systemd hide `/home` and `/tmp`. Fixes #1297.
- Don't set `NoNewPrivileges`, as that breaks some PAM, SELinux, and extauth setups. Fixes #1281 and Debian bug [#836711](https://bugs.debian.org/836711). Also fixes #1347. Also fixes #1376.
- Let systemd stop ejabberd gracefully.
- Let systemd restart ejabberd on failure.
- Don't expose `ejabberdctl reload-config` as an `ExecReload` command, as it doesn't work the way most admins would expect.
- Specify `Type=forking`, as ejabberd is not a `oneshot` command. With `Type=oneshot`, if ejabberd quits for some reason, it cannot be restarted using `systemctl start ejabberd`.
- Increase the file descriptor limit.


mathiasertl:
Thanks for this PR, this causes the default systemd file to actually work. Partly based on what we have [here](https://github.com/jabber-at/ejabberd/blob/master/debian/patches/ejabberd.service.template.patch), my comments:

```diff
 [Unit]
 Description=XMPP Server
-After=network.target
+After=epmd.service network.target
+Requires=epmd.service
```

Also add to the `[Service]` section:

```diff
+ExecStop=/bin/sh -c '@ctlscriptpath@/ejabberdctl reload-config
+PIDFile=/run/ejabberd/ejabberd.pid
```

weiss:
> `After=epmd.service network.target`
> `Requires=epmd.service`

This assumes that a seperate `epmd.service` exists.  It makes sense for Linux distributions to set things up this way and modify the systemd unit file accordingly, but we shouldn't assume that upstream, as epmd(1) comes bundled with Erlang/OTP and is started/stopped implicitly with ejabberd if it's not also used by other Erlang software.

> `ExecStop=/bin/sh -c '@ctlscriptpath@/ejabberdctl reload-config`

I guess you mean `ExecReload` rather than `ExecStop`?  I [removed][1] that intentionally in this PR, because it confuses admins (large parts of ejabberd won't notice that the configuration was reloaded if you don't take further action).  And I don't see a reason to not just call `ejabberdctl reload-config` if you know what you're doing.

> `PIDFile=/run/ejabberd/ejabberd.pid`

This depends on a corresponding `EJABBERD_PID_PATH` in your `ejabberdctl.cfg`.  Same as above: It makes sense for distributions to set things up this way, but we shouldn't assume this upstream.  (Or if we want it that way, we should extend the build magic to also set `EJABBERD_PID_PATH`, but that should probably go into a separate PR.)

[1]: https://github.com/processone/ejabberd/pull/1346/commits/0a3fcc9adea3db4c92bbb2a6d9d2b8ff5b8c8675

PR #1349, opened 2016-10-21
Fix: Replace erlang function with fail-safe jlib function.

prasadvaidya:
As reported in:
 https://www.ejabberd.im/forum/28580/erlang-function-integertobinary1-throwing-badargs-exception

This change is in continuation of earlier [commit](https://github.com/processone/ejabberd/commit/2ab72bcd009536a6bef2be04c71470f17e19d0b2) to handle data type conversion with fail-safe mechanism. Please refer above forum link for details.

This has been verified locally.

### DIFF
--- a/src/ejabberd_sql.erl
+++ b/src/ejabberd_sql.erl
@@ -629,7 +629,7 @@ generic_sql_query_format(SQLQuery) ->
 
 generic_escape() ->
     #sql_escape{string = fun(X) -> <<"'", (escape(X))/binary, "'">> end,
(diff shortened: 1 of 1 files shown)


badlop:
But are those replacements caused by a real posibility that X may be integer/binary?
At least in this case, it is obvious that the code is correct; "I" can only be an integer:
> Args = [<<$$, (integer_to_binary(I))/binary>> || I <- lists:seq(1, N)],

prasadvaidya:
As you pointed out, using fail-safe method would be redundant in this particular case. I have rolled back that particular change.

Issue #1327, opened 2016-10-07
Allow writing custom modules in parent app when ejabberd embedded using elixir


ukutaht:
> What version of ejabberd are you using?

16.08

> What operating system (version) are you using?

MacOS Sierra

> How did you install ejabberd (source, package, distribution)?

Mix Dependency

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

I've created a github repo to exemplify the desired behaviour:
https://github.com/ukutaht/example_ejabberd_mix_module

 It would be neat if users could write custom modules for ejabberd in the same project when it is embedded using mix. The current mechanism to use external modules in `ext_mod` does not seem sufficient for this use case but I may be wrong.
### Implementation notes

Currently the list of available modules is a concatenation of `ext_mod:installed()` and `application:get_key(ejabberd, modules)` ([source](https://github.com/processone/ejabberd/blob/master/src/ejabberd_config.erl#L917))

This list will not include modules written in the parent application. They are neither part of the `ejabberd` application or installed as an external module. However, they are available at runtime so it feels like it should be possible to use them.

I have a branch that solves this by searching through all loaded code at runtime instead of just the `ejabberd` application:
https://github.com/ukutaht/ejabberd/commit/775fd938eaab8ed95595e1fb44b2151259fd4f38

Are there any security implications to doing this? Any better ways to implement this?

zinid:
Your patch will not work, because detecting supported databases of the modules will fail. For example, you don't have `mod_offline_sql` in `code:all_loaded()` list, so ejabberd knows nothing about `sql` database for `mod_offline`.

ukutaht:
@zinid You're right, I gave it a try and it failed to find `mod_offline_sql`

Would you agree, though, that the example I've provided should work? It seems that embedding ejabberd is a fairly new use-case and I wonder if loading modules from other erlang applications is aligned with the vision.

If you think it's a good idea, I'd be happy to put some more effort into finding a good solution.

zinid:
Of course it's a good idea. It's just at the moment I was writing this "db_type by suffix search" the only stuff came to my head was application:get_key(ejabberd, modules).

cromain:
using ejabberd's env as proposed on #1351 is a possible path. we need to check if there is a possible better approach.

cromain:
in ejabberdctl, just add this to EJABBERD_OPTS:
external_beams /path/to/my/beams

[...]

PR #1351, opened 2016-10-23
Allow to add extra modules using application env

kovyl2404:
Hi, guys!

Today I've encountered an issue. I have setup with ejabberd embedded into node with my custom applications. I wrote my own authorization plugin. I was surprised when ejabberd was unable to locate this module, even if it was loaded into the node. I've found that current modules discovery schema requires this module to be placed in [certain place](https://github.com/processone/ejabberd/blob/master/src/ext_mod.erl#L356). It would be cool if I'll be able to add extra modules one more way.

### DIFF
--- a/src/ejabberd_config.erl
+++ b/src/ejabberd_config.erl
@@ -912,8 +912,9 @@ default_db(Host, Module) ->
 
 get_modules_with_options() ->
     {ok, Mods} = application:get_key(ejabberd, modules),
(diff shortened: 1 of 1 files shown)


ukutaht:
Any reason this was closed without merging? I think it would fix #1327

cromain:
hum... i need to check this

cromain:
we'll likely use a similar approach. see #1327 for followup

PR #1353, opened 2016-10-24
Update README

amarjeetkapoor1:
It updates Readme with two requirements that are necessary but not mentioned.

### DIFF
--- a/README
+++ b/README
@@ -106,8 +106,9 @@ To compile ejabberd you need:
 
  - GNU Make.
  - GCC.
+ - autoconf.
  - Libexpat 1.95 or higher.
- - Libyaml 0.1.4 or higher.
+ - Libyaml 0.1.4 or higher (dev).
  - Erlang/OTP 17.1 or higher.
  - OpenSSL 1.0.0 or higher, for STARTTLS, SASL and SSL encryption.
  - Zlib 1.2.3 or higher, for Stream Compression support (XEP-0138). Optional.


weiss:
+ - autoconf.

You'll need Autoconf *if* you compile ejabberd from Git rather than using the [official source tarball][1], yes.

    - - Libyaml 0.1.4 or higher.
    + - Libyaml 0.1.4 or higher (dev).

I guess you had to install some separate `*-devel` package because your Linux distribution splits up Libyaml into multiple packages?  This is distribution-dependent though, and may well happen for other dependencies as well.  It might make more sense to mention this in a separate sentence.

[1]: https://www.process-one.net/en/ejabberd/downloads/

amarjeetkapoor1:
On 24 October 2016 at 20:28, Holger Weiß <notifications@github.com> wrote:
>
> + - autoconf.
>
> You'll need Autoconf if you compile ejabberd from Git rather than using the official source tarball, yes.
>

I think it would make sense that autoconf should be added to
requirements or it can be added where it's mentioned that we don't
need to run ./autoconf if we get the code with configure file.

> - - Libyaml 0.1.4 or higher.
> + - Libyaml 0.1.4 or higher (dev).
>
> I guess you had to install some separate *-devel package because your Linux distribution splits up Libyaml into multiple packages? This is distribution-dependent though, and may well happen for other dependencies as well. It might make more sense to mention this in a separate sentence.
>

This could be added as note but I think it would make sense to mention
in that line only with little more desciption as some people may skip
note or by writting  it like something:
Libyaml 0.1.4 or higher (dev)*.

 and then mention details in note.

Just tell me how exactly it should be or you could make them yourself
as I think these changes are some thing that should be mentioned in
Readme otherwise it wasts 2-3 min to figure them out.

Also there are some syntax error in commands mentioned in
documentation. https://docs.ejabberd.im/admin/ejabberdctl/muc-admin/
Like in

<quote>
ejabberdctl send_direct_invitation room_name password reason jid1[:jid2]
</quote>

It should be:
ejabberdctl send_direct_invitation room_name service password reason jid1[:jid2]

Where could we report that or send corrections?

-- 
Amarjeet Singh
https://amarjeetkapoor1.wordpress.com
https://github.com/amarjeetkapoor1
https://bitbucket.org/amarjeetkapoor

"The journey of a thousand commit begins with a single init"

weiss:
> This could be added as note but I think it would make sense to mention in that line only with little more desciption

I don't really care, but either way it makes no sense to add such a note for Libyaml but not for Libexpat, OpenSSL, Zlib, PAM, and GNU Iconv.

weiss:
> Where could we report that or send corrections?

You can submit documentation corrections against the [processone/docs.ejabberd.im repository][1].

[1]: https://github.com/processone/docs.ejabberd.im

cromain:
100% agree with Holger.
You need autoconf only if you're a developer compiling from git, and in this case you have autoconf anyway, and know you have to use it.
Development packets are distribution-dependent.
I added a minor note anyway.

PR #1358, opened 2016-10-27
Populate included applications to ease embedding

kovyl2404:
I'm building system which runs ejabberd in embeded mode in my custom nodes. In order to ease embedding it would be cool to populate `included_applications` key of `ejabberd` application based on configuration provided by `./configure`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,12 @@ else
 fi
 ])
 
+read -r -d '' INCLUDED_APPS <<'HERE'
+sasl, lager, mnesia, p1_utils, cache_tab,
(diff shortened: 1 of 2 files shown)


badlop:
Hi, can you take a look at the Travis log? It reports some problem, I wonder if it's provoked by your changes or comes from somewhere else:
```
ejabberd_SUITE:init_per_suite failed
Reason: {badmatch,{error,{"no such file or directory","p1_xmlrpc.app...}
```

kovyl2404:
Oh, sorry! It seems my bad. I'll fix this at the week as soon.

PR #1388, opened 2016-11-21
Remove bashismes from ejabberdctl

joudinet:
> We are open to contributions for ejabberd, as GitHub pull requests (PR).
> Here are a few points to consider before submitting your PR. (You can
> remove the whole text after reading.)
> 
> 1. Does this PR address an issue? Please reference it in the PR
>    description.
> 
> 2. Have you properly described the proposed change?
> 
> 3. Please make sure the change is atomic and does only touch the needed
>    modules. If you have other changes/fixes to provide, please submit
>    them as separate PRs.
> 
> 4. If your change or new feature involves storage backends, did you make
>    sure your change works with all backends?
> 
> 5. Do you provide tests? How can we check the behavior of the code?
> 
> 6. Did you consider documentation changes in the
>    processone/docs.ejabberd.im repository?

To avoid unecessary extra quoting, do not call commands with sh -c ''
and use the -- option from su to supply arguments.

Parse command line parameters is a bit tricky as the previous behavior
allows to mix options to ejabberdctl with unknown options given to
the next script (usually, the ctl). This is solved by relying on the
fact that for loop saves its argument, so we can flush its content
with set -- and re-add unknown options with set -- "$@" "$arg".

Finally, remove unecessary quotes in mnesia options and in the ping
command.


joudinet:
I'm afraid I don't understand these errors and it looks like no recent PRs succeed either. Between the two coverage tests, I've simply merged with the last changes in master, so I can't explain the coverage decrease. For the travis test suite, it fails for OTP 17.5 and OTP 19.1 but succeeds for OTP 18.3.
Any idea on how to solve these issues?

zinid:
@joudinet tests are now being refactored, so don't pay attention.

cromain:
@joudinet I merged your changes, and refactored the entire script, to make it smarter and simpler

PR #1393, opened 2016-11-25
Fixes for pt-br translations

rodrigues:
🇧🇷👌


badlop:
I've applied your fixes to the file pt-br.po, which is the source file. Next time make your changes to the pt-br.po file.

PR #1395, opened 2016-11-27
Use -set-cookie when running erl in ejabberdctl.template.

bowlofeggs:
This patch converts the ejabberdctl.template to use erl's
-set-cookie option when an .erlang.cookie file is found in
$SPOOL_DIR. This will help for installations where ejabberd might
be started by one user, but ejabberdctl might be used by a
different user to manage it. For example, in Fedora ejabberd is
launched by systemd as the ejabberd user, but users might want to
use "sudo ejabberdctl" to manage the node.


bowlofeggs:
I learned that the erlang cookie is a bit like a password (I had thought it was just the id of the node the user wants to connect to). Thus it should not be passed with ``-set-cookie`` because we don't want it to appear in the process table.

PR #1446, opened 2017-01-03
Use include_lib() to import external libraries.

bowlofeggs:
This commit adjusts several ```include()``` statements across the project to use ```include_lib()``` instead so that the libraries can be found from a system install. This commit enables ejabberd-16.12 to be built in Fedora Rawhide with ```xmpp``` and ```fast_xml``` being installed in ```/usr/lib64/erlang/lib/```.

If there's a better way to do this, I'd love to know about it (I'm new to Erlang)!

### DIFF
--- a/include/jlib.hrl
+++ b/include/jlib.hrl
@@ -18,8 +18,8 @@
 %%%
 %%%----------------------------------------------------------------------
 
--include("ns.hrl").
--include("fxml.hrl").
+-include_lib("xmpp/include/ns.hrl").
+-include_lib("fast_xml/include/fxml.hrl").
 
 -define(STANZA_ERROR(Code, Type, Condition),
 	#xmlel{name = <<"error">>,
--- a/src/acl.erl
+++ b/src/acl.erl
@@ -41,7 +41,7 @@
 
 -include("ejabberd.hrl").
 -include("logger.hrl").
--include("jid.hrl").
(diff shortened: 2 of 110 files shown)


bowlofeggs:
See also https://github.com/processone/xmpp/pull/3 where I had a similar problem when trying to build xmpp in Fedora Rawhide. (Please do educate me if there's a better way for me to get the build to find those includes!)

bowlofeggs:
I am not able to agree to the contributor license agreement. I do, however, agree to the [DCO](https://developercertificate.org/) if that helps, since it does not grant my contribution under a different license than the project is granted under.

bowlofeggs:
@mremond though I can't sign the agreement, I do think it's pretty reasonable to say that the change I proposed here is trivial enough that it wouldn't be a problem if someone else just did the same thing. I would like to see the change or an equivalent change go in. It's not like I invented time travel here ☺

bowlofeggs:
(otherwise I have to keep carrying this patch in Fedora, which isn't a huge problem, but I'd prefer if Fedora could stay as close to upstream as possible without having to maintain extra patches…)

mremond:
@bowlofeggs Thanks !

And many thanks for your work on the Fedora package. We would love to be able to make it easier to build for distribution maintainers.

PR #1499, opened 2017-01-27
Include stanza ID with archived offline messages

### LINKED ISSUE #1480: Missing stanza-id in Message (from offline storage)

17.01.45

Ubuntu Linux

source

I'm receiving a message without a stanza-id. It might be from the offline storage.

```xml
<message xmlns="jabber:client" to="romeo@example.com" from="juliet@example.com" type="chat" id="402466435698">
    <delay xmlns="urn:xmpp:delay" from="example.com" stamp="2017-01-23T16:37:22.445939Z">Offline storage</delay>
    <body>...</body>
</message>
```

To not having duplicate messages in the local client archive, the stanza-is are needed in all messages.

weiss:
Fixes #1480.

### DIFF
--- a/src/ejabberd_piefxis.erl
+++ b/src/ejabberd_piefxis.erl
@@ -526,8 +526,7 @@ process_offline_msg(#message{from = undefined}, _State) ->
 process_offline_msg(Msg, State = #state{user = U, server = S}) ->
     To = jid:make(U, S),
     ejabberd_hooks:run_fold(
-      offline_message_hook, To#jid.lserver, pass,
-      [xmpp:set_to(Msg, To)]),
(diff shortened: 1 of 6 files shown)


zinid:
We probably don't need to pass `From` and `To` anymore, as it's a part of `#message{}`. Since we decided to break the hook's backward compatibility, it is now a good moment doing this (I already did that in a lot of hooks).

weiss:
Ah, okay.  I'll update the patch.

weiss:
I've updated the patch to make it work with the current code (without separate `From` and `To` arguments).

zinid:
Thanks :smiley_cat:

PR #1548, opened 2017-02-16
Use ejabberd_config:get_option/2 instead of ejabberd_config:get_local_option/2

lemenkov:
All functions from ejabberd_config:get_{global|local}_option/X family just calls ejabberd_config:get_option of the same arity. So I believe it's better to use get_option everywhere (in two remaining places).

Should we remove exported get_local_option/get_global_option or at least mark them as deprecated?

Also add_local_option and add_global_option functions also just calls add_option. So they should be treated in the same way.

### DIFF
--- a/src/ejabberd_auth_external.erl
+++ b/src/ejabberd_auth_external.erl
@@ -315,7 +315,7 @@ get_mod_last_configured(Server) ->
     end.
 
 is_configured(Host, Module) ->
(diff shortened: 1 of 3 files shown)


zinid:
I think we can deprecate them indeed.
I wouldn't remove those functions because it's not a problem for us to keep them. We only remove functions when it's significant burden for us to keep backward compatibility.

PR #1555, opened 2017-02-19
Let ejabberd_c2s handle <session/> IQ

weiss:
Let the stream management code see `<session/>` IQs.  This fixes the incoming stanza count.

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -448,6 +448,13 @@ handle_authenticated_packet(Pkt, #{lserver := LServer, jid := JID,
     case Pkt2 of
 	drop ->
 	    State2;
+	#iq{type = set, sub_els = [_]} ->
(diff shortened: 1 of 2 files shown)


zinid:
Thanks.

zinid:
There is a mess with `Pkt1` <-> `Pkt2`. I think we should use `Pkt2` in these new branches you added.

zinid:
Otherwise privacy lists don't work (the 'active' privacy list is not set in `Pkt1`'s meta).

weiss:
Oh, I created the patch before you [introduced][1] `Pkt2` :-)  And then overlooked the change, sorry.  [Fixed][2].

[1]: https://github.com/processone/ejabberd/commit/6fd89f5fe0ecb39d1e8f6b60bb403d03cbf92582
[2]: https://github.com/processone/ejabberd/commit/61cdee97fc1f808741b5950c77b7e1d9ca5fd611

Issue #1586, opened 2017-03-03
Support for GSSAPI authentication


badlop:
There was a long-standing feature request in the old JIRA bug tracking system:
https://support.process-one.net/browse/EJAB-831 GSSAPI authentication

I attach here the latest working patch and a backup of the conversation.
[831-fixed.patch.txt](https://github.com/processone/ejabberd/files/816885/831-fixed.patch.txt)
[831-conversation.txt](https://github.com/processone/ejabberd/files/816891/831-conversation.txt)

This patch was not included in ejabberd main repository because nobody at ProcessOne knows enough about GSSAPI to review and maintain the code, and apparently no known collaborator either.

If at least a tester were available, it would be interesting to try to rework the code so most of it can be committed to ejabberd-contrib, and the barely minimum to ejabberd main repo.

debalance:
Thanks @badlop for bringing this up again!
I'll contact the people from the according wishlist bug in Debian (https://bugs.debian.org/590503) and direct them here.

debalance:
@badlop, I cannot build your fork of esasl on Debian sid (using automake 1.15):

esasl$ ./bootstrap 
configure.ac:27: installing 'config/install-sh'
configure.ac:27: installing 'config/missing'
configure.ac:51: error: required file 'esasl/Makefile.in' not found
configure.ac:51: error: required file 'esasl/priv/Makefile.in' not found
configure.ac:51: error: required file 'esasl/src/Makefile.in' not found

esasl$ automake -a -c
configure.ac:51: error: required file 'esasl/Makefile.in' not found
configure.ac:51: error: required file 'esasl/priv/Makefile.in' not found
configure.ac:51: error: required file 'esasl/src/Makefile.in' not found

lemenkov:
Hello! I've got a working patch-set based on this one (we've tested it very recently in a real Kerberos infrastructure):

https://github.com/lemenkov/ejabberd/commits/fedora-16.12-gssapi-test

I can squash them down and make a PR.

samweisgamdschie:
Hi there!

Is there a way to add another efforts to implement this awesome feature?

We will really appreciate it!

Neustradamus:
I would like news too.

Neustradamus:
@dequbed has added a PR: https://github.com/processone/xmpp/pull/56: WIP, thanks!

PR #1595, opened 2017-03-06
Support SASL GSSAPI authentication

lemenkov:
This PR is based on patchset floating around in internet since 2009.

Patchset since then was improved by @badlop, Fyodor Kupchik (https://launchpad.net/~ferimy), Konstantin L. Metlov (https://launchpad.net/~metlov), and many others.

This particular version was tested against working Kerberos installation at Fedora Project. Thanks for @bowlofeggs and @puiterwijk for that!

For the record - this is a patchset's breakdown:

* https://github.com/lemenkov/ejabberd/compare/3114f5c...75c1389
* https://github.com/lemenkov/ejabberd/tree/fedora-16.12-gssapi-test

This PR closes bug #1586.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,7 +101,7 @@ AC_ARG_ENABLE(mssql,
 esac],[db_type=generic])
 
 AC_ARG_ENABLE(all,
-[AC_HELP_STRING([--enable-all], [same as --enable-odbc --enable-mysql --enable-pgsql --enable-sqlite --enable-pam --enable-zlib --enable-riak --enable-redis --enable-elixir --enable-iconv --enable-debug --enable-tools (useful for Dialyzer checks, default: no)])],
+[AC_HELP_STRING([--enable-all], [same as --enable-odbc --enable-mysql --enable-pgsql --enable-sqlite --enable-pam --enable-zlib --enable-riak --enable-redis --enable-elixir --enable-esasl --enable-iconv --enable-debug --enable-tools (useful for Dialyzer checks, default: no)])],
 [case "${enableval}" in
   yes) odbc=true mysql=true pgsql=true sqlite=true pam=true zlib=true riak=true redis=true elixir=true iconv=true debug=true tools=true ;;
   no) odbc=false mysql=false pgsql=false sqlite=false pam=false zlib=false riak=false redis=false elixir=false iconv=false debug=false tools=false ;;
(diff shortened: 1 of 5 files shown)


zinid:
Several issuer I found quickly:
1) `try_start_gssapi()` should be called from inside cyrsasl_gssapi:start/1, obviously;
2) Attempt to attach directly to `ejabberd_sup`: `master` won't even start with this, it will get stuck in deadlock when starting `cyrsasl` gen_server;
3) Why we need to attach some third-party software (`esasl`) to local supervisor is beyond me: this suggests `esasl` itself has design flaws;
4) We cannot return `{error, <<"not-authorized">>}` or something like that in new ejabberd API
5) Auth module is not included in `{ok, [{username, Authid}, {authzid, Authzid}]}`.

Note these are problems I detected quickly in a couple of seconds reading the module, I think I can find more with deep reading.

lemenkov:
@zinid thanks for the quick answer!

Regarding esasl - indeed it doesn't have optimal design. We have to use it because "it just works". Later we should replace it with a better solution (yet to be written).

zinid:
The software with such design cannot "just work". I looked inside it and it looks like shit.
You see, if something goes wrong and a user will report a bug, who will handle this?

lemenkov:
@zinid it actually can work, so no need to throw such rhetorical questions :). We plan to rewrite it later. We kept it as is just to focus on actual testing of this feature (otherwise there will be too much moving parts). So it will be rewritten eventually.

Just for the record - I see this PR mostly as a RFE, not yet ready for inclusion. And I'm happy to receive any comments and critics.

zinid:
I'm acutally fine with adding normal OTP structure to `esasl` and removing at least those glitches I have already mentioned.

lemenkov:
@zinid I did a few restructuring changes here:

https://github.com/lemenkov/esasl/tree/rebarize

At least it builds with rebar now, although codebase is still the same.

Neustradamus:
Any news?
We arrive in 2019 ^^

sim0nx:
Any news on this ? :-)

zinid:
None of the issues I mentioned are fixed, so no, no news.

Neustradamus:
@lemenkov: Any news on it?

lemenkov:
> @lemenkov: Any news on it?

Nope, unfortunately. I am no longer using XMPP in any way.

licaon-kter:
@lemenkov /close this then

Neustradamus:
@licaon-kter: The code is good if it is not perfect, one people can update it...

licaon-kter:
@Neustradamus the code is almost 4 years old, the coder didn't even show interest all these years and doesn't care anymore. _Let it go_

Neustradamus:
@dequbed has added a PR: https://github.com/processone/xmpp/pull/56: WIP, thanks!

PR #1603, opened 2017-03-11
Note how to apply s2s Access Rights in the example

strugee:
This bit me in production. Took a long while to debug why my spam policies weren't working, so I thought I'd share the knowledge. A big problem here is that while a lot of the commented-out options are defaults, the referenced line is not.

I'm not sure this is the best way to do this - I'm happy to change the PR if folks want. Other ways to do this that I considered:

1. Uncomment the relevant line by default
2. Make `s2s_access: s2s` the default

Both of those feel magical, though.


strugee:
Thanks! \o/

PR #1620, opened 2017-03-18
Fix for network.target dependency

radantuhy:
Since the network.target is actually a requirement, it should be stated in "Requires=" as well. If missed, the system (tested on CentOS7) will attempt to stop the service after networking stops which doesn't work and causes unnecessary delay with a default (typically 90sec) timeout. Please note that the "After=" statement has to be there as well.

### DIFF
--- a/ejabberd.service.template
+++ b/ejabberd.service.template
@@ -1,5 +1,6 @@
 [Unit]
 Description=XMPP Server
+Requires=network.target
 After=network.target
 
 [Service]


weiss:
> Since the network.target is actually a requirement, it should be stated in "Requires=" as well. If missed, the system (tested on CentOS7) will attempt to stop the service after networking [...].

It won't.  The existing `After=` statement already tells systemd to stop the `ejabberd.service` before the `network.target` if both are running, and the [systemd.special(7)][1] man page advises against adding a `Requires=` line:

> This unit is supposed to indicate when network functionality is available, but it is only very weakly defined what that is supposed to mean, with one exception: at shutdown, a unit that is ordered after network.target will be stopped before the network — to whatever level it might be set up then — is shut down. It is hence useful when writing service files that require network access on shutdown, which should order themselves after this target, but not pull it in.

So if the shutdown isn't clean on your system, you should have to take a closer look at what exactly is going wrong.  Depending on the details of your system, it might help to depend on the `network-online` target instead (see the [systemd.special(7)][1] man page or [this article][2]).  If it does, please let us know.

[1]: https://www.freedesktop.org/software/systemd/man/systemd.special.html
[2]: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

weiss:
> So if the shutdown isn't clean on your system, you should take a closer look at what exactly is going wrong.  Depending on the details of your system, it might help to depend on the `network-online` target instead (see the [systemd.special(7)][1] man page or [this article][2]).  If it does, please let us know.

As we got no additional feedback, I'll close this PR.  Feel free to open a new issue (referencing this one) if you want to provide further information.

[1]: https://www.freedesktop.org/software/systemd/man/systemd.special.html
[2]: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

PR #1660, opened 2017-04-06
Fix order of starting supervisors.

michal800106:
ExtMod should be loaded before GenModSupervisor because ext_mod adds proper paths to ebin from modules (ejabberd-contrib).
Without this change you have to add -pa parameter with path to module's ebin.

### DIFF
--- a/src/ejabberd_sup.erl
+++ b/src/ejabberd_sup.erl
@@ -173,7 +173,7 @@ init([]) ->
 	   RouterMulticast,
 	   Local,
 	   SM,
-	   GenModSupervisor,
 	   ExtMod,
+	   GenModSupervisor,
 	   Auth,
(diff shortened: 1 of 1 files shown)


zinid:
In this case there is no way to use `gen_mod:start_child()` in external modules, because `ejabberd_gen_mod_sup` is not started yet.

michal800106:
Ok. But in this way there is cycle dependency between ext_mod and gen_mod
ext_mod tries to config modules but now before ext_mod is launched gen_mod.

@zinid  Are you sure that gen_mod:start_child is launched from ext_mod? Because I see:
https://github.com/processone/ejabberd/blob/master/src/gen_mod.erl#L195

And in my opinion ext_mod is only preparing environment for modules. It doesn't start modules itself.

PR #1664, opened 2017-04-06
Fix routing of groupchat and headline messages

weiss:
As per [the spec][1],

* silently drop headline messages sent to the bare JID of an offline user or to the full JID of an unavailable resource, and
*  don't (re)route groupchat messages sent to the bare JID or to unavailable resources, but generate an error instead.  This avoids duplicated MUC messages.

[1]: https://tools.ietf.org/html/rfc6121#section-8


zinid:
Looks fine

PR #1677, opened 2017-04-17
lager_crash_log in some cases not run, catch it

getong:
In function `do_start_for_logger`, the `crash_log` is set to `false`.  In this situation, `lager_crash_log` not runs, catch it.


zinid:
Thanks ;)

Issue #1695, opened 2017-04-24
Building the docker image with 17.04 fails


pipo02mix:
> What version of ejabberd are you using?
17.04
> What operating system (version) are you using?
Mac Os Sierra 10.12.4
> How did you install ejabberd (source, package, distribution)?
Dockerfile
> What did not work as expected? Are there error messages in the log? What
-----
Writing ssl cert and private key to '/opt/ejabberd/ssl/localhost.pem'...
Generating ejabberd config file...
Generating ejabberdctl config file...
echo "Starting ejabberd..."
/sbin/run: line 92: /sbin/ejabberdctl: No such file or directory
/sbin/run: line 91: /sbin/ejabberdctl: No such file or directory
The command '/bin/sh -c run configure' returned a non-zero code: 1
> was the unexpected behavior? What was the expected result?

ejabberdctl is found and the image can be built.

What I did is change the tag version on the Dockefile because for the branch/tag 17.04 still on 17.03. (For 17.03 the image is built without problem) But with new tag ejabberd fails when it is started because ejabberdctl looks is not there. Any hint?

imhotepisinvisible:
The default installation prefix changed in this release, so you need to edit the configure line in the Docker file to include `prefix=/`

pipo02mix:
Thanks @imhotepisinvisible, it works.

PR #1697, opened 2017-04-25
Update Dockerfile to be able to build 17.04 ejabberd version

pipo02mix:
Because the default installation prefix has changed in the new release, it is needed to add in the configure line of Docker file the argument  --prefix=/

Issue related https://github.com/processone/ejabberd/issues/1695


cromain:
indeed, thanks for PR

Issue #958, opened 2016-02-17
ejabberd 14 breaks ejabberd_ munin plugin


PF4Public:
Since ejabberd 14 installation `ejabberd_` munin plugin stopped working.
I'll cross-link two issues
[munin bugreport](https://github.com/munin-monitoring/munin/issues/488)

Any help in resolving is greatly appreciated.

mremond:
@kolargol Do you think this issue still valid ?

PF4Public:
This is still broken with ejabberd 16.03-1

mremond:
Well, I think the change will have to be made on munin plugin, as I do not think the issue is in ejabberd. From the ticket you link it seems unrelated to ejabberd.

PF4Public:
I believe it has something to do with both munin and ejabberdctl:

> Still I believe that root cause is not setting gid by munin. At least it could be the first obstacle.
 Should I fool  ejabberdctl  by replacing  -g  with  -u  in  id  checks it no longer complains and seems to pass given arguments forward to  erl , however this still fails with a bunch of access denied
...
Therefore I think that gid check ( -g ) in  ejabberdctl  was made on purpose.

rraptorr:
Why don't you simply configure munin to change group to ejabberd?

PF4Public:
@rraptorr Because it does not change a thing.
Just tried to configure munin with root user and root group and it worked. But not with ejabberd user/group. Using cli with ejabberd user however still works fine.

rraptorr:
In your other bug report you indicate that you put 'user ejabberd' in munin plugin config. There is no mention if you also put 'group ejabberd' in that config. Have you? If you put 'group ejabberd' and it still does not set it, then it is clearly a bug in munin.

PF4Public:
Consider the code

> echo $(whoami)
echo
echo $(id -G)
echo
echo $(id -g)
echo
echo $(id -g ejabberd)
echo
echo $(id -g munin)

If I put 

> user root
group root

I get proper statistics and

> root
139 0
139
123
139

Otherwise if I put 

> user ejabberd
group ejabberd 

I got no statistics and 

> ejabberd
139 123
139
123
139

So, what is the difference here and why it works with root and does not with ejabberd?

rraptorr:
I have no idea, ask munin guys as it really seems like some problem with munin and its handling of user/group. Especially since ejabberdctl with sudo works.

PF4Public:
OK, I've got it.
[...]

PR #1716, opened 2017-05-06
ejabberdctl: allow execution by 'root' user

sumpfralle:
previously the code checked the following conditions:
* is the user a member of the "root" group (not necessarily as its primary group)
    * effect: use `su`
* is the primary group of the user the same as the primary group of INSTALLUSER
    * effect: just user `bash -c`

The previous behaviour contradicts the resulting error message (if none
of the above is valid):
```
This command can only be run by root or the user $INSTALLUSER"
```

This change additionally checks for the following condition:
* is the user "root"
    * effect: use `su`

This fixes the issue #958 (by allowing to use `user root` in the munin plugin configuration).

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -20,20 +20,29 @@ ERL_LIBS={{libdir}}
 # check the proper system user is used if defined
 if [ "$INSTALLUSER" != "" ] ; then
     EXEC_CMD="false"
+    # is the user part of the group "root"?
     for GID in `id -G`; do
         if [ $GID -eq 0 ] ; then
-            INSTALLUSER_HOME=$(getent passwd "$INSTALLUSER" | cut -d: -f6)
-            if [ -n "$INSTALLUSER_HOME" ] && [ ! -d "$INSTALLUSER_HOME" ] ; then
-                mkdir -p "$INSTALLUSER_HOME"
-                chown "$INSTALLUSER" "$INSTALLUSER_HOME"
-            fi
-            EXEC_CMD="su $INSTALLUSER -c"
+            EXEC_CMD="become_root"
         fi
     done
+    # is the user "root"?
+    if [ "$(id -u)" -eq 0 ] ; then
+        EXEC_CMD="become_root"
+    fi
+    # a properly set primary group is sufficient (no need to switch identities)
     if [ `id -g` -eq `id -g $INSTALLUSER` ] ; then
         EXEC_CMD="bash -c"
     fi
-    if [ "$EXEC_CMD" = "false" ] ; then
+    if [ "$EXEC_CMD" = "become_root" ] ; then
+        # prepare an empty install user's home directory, if it is missing
+        INSTALLUSER_HOME=$(getent passwd "$INSTALLUSER" | cut -d: -f6)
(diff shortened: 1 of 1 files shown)


sumpfralle:
Regarding the error message (`This command can only be run by root or the user $INSTALLUSER"`): even with above change it is still not possible to run `ejabberdctl` with the INSTALLUSER (but without its primary group). This can happen if non-interactive programs try to gain appropriate privileges for starting ejabberdctl. For example the ejabberd munin plugin in its current state causes the following set of user attributes:
```shell
$ id
uid=105(ejabberd) gid=105(munin) groups=105(munin)
```
According to the error message in `ejabberdctl` the above state should be sufficient (we are the INSTALLUSER).

Anyway: I am not sure why the checks in `ejabberdctl` focus on the group (instead of the user). Thus maybe it is technically appropriate to enforce the primary group.

cromain:
we can't merge this now, but i think this is covered by new version of ejabberdctl
I assign this to 17.08 so we can check before closing.

sumpfralle:
Thank you for taking a look!

Sadly the current state will still fail to detect the user root - which should be really the most trivial thing to test (`[ "$(id -u)" -eq 0 ]`). The current state of ejabberctl checks for the membership of the 'wheel' group (which is weird, since that group has no special powers by definition - as far as I understand). Maybe there are systems out there, where the user root is member of the 'wheel' group - but at least here on Debian the 'wheel' group does not even exist at all.

Thus third-party tools like the munin-plugin will still have a hard time to use ejabberdctl. This is a pity.

I would be happy to prepare a new changeset, if I would at least know, what the "group" checks are supposed to do. Why don't we check for specific permissions (e.g. being able to access certain sockets). Why do we check group memberships instead of user IDs?

Thank you for your time!

cromain:
The wheel group exists to lists people who are able to su to root. But it seems less and less used in Linux world, so we have to improve this.
We must indeed add the simple user test by uid, and maybe drop the group check.
this will fallback to this:
 if ejabberdctl run as root then
    if install_user defined then EXEC_CMD=as_install_user
    else EXEC_CMD=as_current_user (and display warning this is not good idea to run as root)
 else if ejabberdctl run as install_user then
     EXEC_CMD=as_current_user
 else
    quit with error

for any other use, it will be up to caller to 'su ejabberd' itself and be allowed on the system to do so.

I think we can go this way.

sumpfralle:
Thank you for summarizing the path to a better future :)

Since you assigned this pull request to yourself: do you want to fix this on your own or would you like me to prepare a new pull request?

cromain:
i'll fix it, taking into account other current PR around this topic.
thanks.

cromain:
I have simplified the user check, as described previously.

sumpfralle:
Thank you!

I have two nitpicks, but they are probably not worth the effort:
* the username `root` is not magic - but the UID zero is the real thing, that turns a user into a privileged one - but probably this is not relevant
* you used the old command substitution (backquotes) instead of `$(...)`

(just tell me, if you think it is worth a pull request)

Anyway: thank you for making it possible again to run the munin plugin for ejabberd.

PR #1730, opened 2017-05-16
Allow c2s_terminated to know reason as shutdown

nOmi7:

   description.

   them as separate PRs.

### DIFF
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -47,9 +47,9 @@
 	 process_terminated/2, process_info/2]).
 %% API
 -export([get_presence/1, get_subscription/2, get_subscribed/1,
-	 open_session/1, call/3, send/2, close/1, close/2, stop/1,
-	 reply/2, copy_state/2, set_timeout/2, route/2,
-	 host_up/1, host_down/1]).
+	 open_session/1, call/3, send/2, close/1, close/2, 
+	 stop/1, shutdown/1, reply/2, copy_state/2, set_timeout/2, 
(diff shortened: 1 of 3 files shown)


zinid:
https://github.com/processone/ejabberd/pull/1730/files#diff-4da3143d0c53219913b816c23d48feb7R131

LOL :D

zinid:
I also think this will not always work, because a c2s process may terminate faster than `ejabberd_c2s:shutdown/1` is called (since stream errors always lead to termination).

zinid:
To be clear:
```erlang
ejabberd_c2s:send(Pid, xmpp:serr_system_shutdown()),
%% Scheduler switched task
%% Pid terminated with reason `normal`
%% Scheduler switched task back
%% shutdown is casted to non-existing process Pid
ejabberd_c2s:shutdown(Pid);
```

nOmi7:
What needs to be done for it work properly then :/ i wish i could be like you :/ please guide me to correct it if its acceptable

zinid:
Well, using some atomicity? Like not sending system-shutdown, but 'shutdown' event should lead to sending it with proper stop reason after.

zinid:
Something like
```erlang
handle_cast(shutdown, State) ->
    {stop, shutdown, send_err_system_shutdown(State)}.
```
Note: this is **NOT** ready to use code

zinid:
Also, if you implement this in c2s, please take care about s2s too.

nOmi7:
aye aye captain thanks for help will correct it

Issue #1738, opened 2017-05-19
is it possible to update module spec from any github repo?


HeroPD:
> What version of ejabberd are you using?

17.04

> What operating system (version) are you using?

OSX

> How did you install ejabberd (source, package, distribution)?

source

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

badlop:
No, the repository is specified in ext_mod.erl, not configurable

cromain:
ext_mod handles ejabberd-contrib repos.
but you can add your repository using add_sources/2

ext_mod:add_sources(my_mod, "https://github.com/me/my_mod").

PR #1740, opened 2017-05-20
update spec from custom github repo ext_mod.erl

HeroPD:
issue #1738

i modified ext_mod.erl little bit

i created repo for test case
ejabberdctl modules_update_specs_from_github https://github.com/HeroPD/ejabberd_module_test

### DIFF
--- a/src/ext_mod.erl
+++ b/src/ext_mod.erl
@@ -85,6 +85,13 @@ get_commands_spec() ->
                         module = ?MODULE, function = update,
                         args = [],
                         result = {res, integer}},
+     #ejabberd_commands{name = modules_update_specs_from_github,
+                        tags = [admin,modules],
+                        desc = "",
+                        longdesc = "",
+                        module = ?MODULE, function = update,
+                        args = [{module, binary}],
+                        result = {res, integer}},
      #ejabberd_commands{name = modules_available,
                         tags = [admin,modules],
                         desc = "",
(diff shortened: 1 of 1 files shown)


cromain:
This PR is not completely needed, as ext_mod already supports extra contrib:
```
(ejabberd@localhost)1> ext_mod:add_sources(mod_test, "https://github.com/HeroPD/ejabberd_module_test").
ok
(ejabberd@localhost)2> ext_mod:install(mod_test).
/home/chris/.ejabberd-modules/sources/ejabberd_module_test/mod_test/src/mod_test.erl:3: Warning: undefined callback function depends/2 (behaviour 'gen_mod')
/home/chris/.ejabberd-modules/sources/ejabberd_module_test/mod_test/src/mod_test.erl:3: Warning: undefined callback function mod_opt_type/1 (behaviour 'gen_mod')
ok
(ejabberd@localhost)3> ext_mod:installed(mod_test).
true
```
The thing is actual implementation lack automatic update of such repository. (the same way as ejabberd-contrib). I'll check this and also your second commit about compile_includes/0

cromain:
I've adopted a simpler approach, allowing custom repository to be updated with ext_mod:update()
contributed modules can also depend at each others (installed module include directory are added to compilation path)

HeroPD:
I have include folders in subdirectory as `deps/*/include` . I am adding them using `code:add_patha/1` so second commit was about adding subdirectory `include` in module compiler that way we don't have to copy every `*.hrl` files

HeroPD:
Does ext mode support rebar?

Issue #1765, opened 2017-06-08
-name : Unknown option No file named ejabberd@localhost when doing iexlive


IRog:
> What version of ejabberd are you using?

17.04.160

> What operating system (version) are you using?

Ubuntu 17.04

> How did you install ejabberd (source, package, distribution)?

Source

> What did not work as expected? Are there error messages in the log? What

`./ejabberdctl iexlive`

> was the unexpected behavior? What was the expected result?
Starting a ejabberd with a iex live shell attached to it

I was also having problems with `iexdebug`
This might be related to my installation. I built it from source and followed the elixir guide to build it in $HOME/my-ejabberd I also did this as a user called ejabberd. Is this something wrong with mnesia or the erlang cookie? `./ejabberdctl live` works...

IRog:
https://github.com/processone/ejabberd/pull/1766

cromain:
good catch, thanks.

PR #1766, opened 2017-06-08
fixing exec_iex bug with proper --name arguement

IRog:
https://github.com/processone/ejabberd/issues/1765 Fix

Fixing `./ejabberctl iexlive` not working because of an incorrectly formed `-name` argument which should have been `--name`

Build from source and then `./ejabberctl iexlive` works


cromain:
indeed, good catch

PR #1793, opened 2017-06-21
Fix mod_muc_admin:set_room_affiliation

marcphilipp:
The following command fails in 17.04:

```
ejabberdctl set_room_affiliation my-room conference.my-domain user-id@my-domain none
```

The root cause is that `send_kickban_presence1/6` in `mod_muc_room` expects `undefined | jid()` as it’s first parameter (`UJID`). However, `mod_muc_admin` passes an empty binary string (last element in tuple):

```
gen_fsm:sync_send_all_state_event(Pid,
    {process_item_change, {jid:decode(JID), affiliation, Affiliation, <<"">>}, <<"">>})
```

Thus, `jid:tolower(UJID)` in `mod_muc_room:send_kickban_presence1/6` fails.

In addition, the root cause is hidden by a bug in the catch block of `process_item_change/3` where `jid:encode(UJID)` is called with the empty binary string.

This pull request changes `process_item_change/3` to also work with `undefined` as `UJID` and passes `undefined` from `mod_muc_admin`.

Unfortunately, I couldn’t find a unit test for `process_item_change`. I’d be happy to contribute one, but didn’t know where to start or where to put it.

### DIFF
--- a/src/mod_muc_admin.erl
+++ b/src/mod_muc_admin.erl
@@ -1012,7 +1012,7 @@ set_room_affiliation(Name, Service, JID, AffiliationString) ->
     case mod_muc:find_online_room(Name, Service) of
 	{ok, Pid} ->
 	    %% Get the PID for the online room so we can get the state of the room
-	    {ok, StateData} = gen_fsm:sync_send_all_state_event(Pid, {process_item_change, {jid:decode(JID), affiliation, Affiliation, <<"">>}, <<"">>}),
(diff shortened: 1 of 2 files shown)


marcphilipp:
Thanks for merging this so quickly! 👍 

How about the test? I think an integration test for `ejabberdctl set_room_affiliation` would make sense, wouldn't it?

zinid:
We do not have a framework for testing ejabberd commands, as far as I'm concerned.

marcphilipp:
Is there a framework for integration test in general?

zinid:
https://github.com/processone/ejabberd/tree/master/test
But it's a bit hairy, mind you.

Issue #1804, opened 2017-06-26
Remove ejabberd_router:is_my_route(From#jid.lserver) check from mod_block_strangers


IRog:
I just wanted to check why this line was one of the criteria to skip filtering. I suppose if they're a user on your system as well they might not be strangers?

This is line 64 of src/mod_block_strangers.erl I just want to check the intention, before submitting a PR to get rid of the line or forking the module with that behavior removed...

IRog:
https://github.com/processone/ejabberd/pull/1809 let me know, if you want me to call it something else or if there are any other problems.

PR #1809, opened 2017-06-27
Adding allow_local_users config option to mod_block_strangers

IRog:
This is addressing https://github.com/processone/ejabberd/issues/1804
I wasn't sure what to call the option, so feel free to pick something new.
I default it to true to maintain the old behavior by default.
I could also update the documentation on the module, if someone shows me how to do that.


alexeyshch:
Thanks for the patch, I've added the option to mod_opt_type and committed it.

IRog:
I could have modified my PR...

PR #1815, opened 2017-06-28
Fix detection of 'wheel' membership and quoting

sumpfralle:
Previously any group containing the string 'wheel' would be sufficient for picking the 'as_install_user' code path.


sumpfralle:
As a side-note: hard-coding the specific behaviour of a 'wheel' group feels wrong. This should be done by the system via sudo or other means.

cromain:
I have simplified the user check, dropping gid matching and opted for a simpler uid only approach as described in #1716. We don't check for wheel group anymore.

PR #1821, opened 2017-07-02
fixing mix compile when using ejabberd as an elixir dependency

IRog:
`mix deps.compile ejabberd` was failing prior to this change. I tested it with the following mix file and doing `mix do deps.get, compile` from within the mix project:

```elixir
defmodule Ejapp.Mixfile do
  use Mix.Project

  def project do
    [app: :ejapp,
     version: "0.1.0",
     elixir: "~> 1.4",
     build_embedded: Mix.env == :prod,
     start_permanent: Mix.env == :prod,
     deps: deps()]
  end

  def application do
    [extra_applications: [:logger, :ejabberd]]
  end

  defp deps do
    [{:ejabberd, git: "https://github.com/IRog/ejabberd.git", branch: "mixCompile"}]
  end
end

```

Without this change using ejabberd as an Elixir mix dependency seems to be broken.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -10,6 +10,7 @@ defmodule Ejabberd.Mixfile do
      compile_path: ".",
      compilers: [:asn1] ++ Mix.compilers,
      erlc_options: erlc_options(),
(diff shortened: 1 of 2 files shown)


IRog:
Rebasing into new branch to do only 1 commit

Issue #1855, opened 2017-07-11
This command can only be run by root or the user ejabberd


jarbelix:
> What version of ejabberd are you using?

dpkg -i ejabberd_17.07-0_amd64.deb

> What operating system (version) are you using?

lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.0 (stretch)
Release:	9.0
Codename:	stretch

> How did you install ejabberd (source, package, distribution)?

https://www.process-one.net/downloads/downloads-action.php?file=/ejabberd/17.07/ejabberd_17.07-0_amd64.deb

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

/opt/ejabberd-17.07/bin/start 
ERROR: This command can only be run by root or the user ejabberd
ERROR: This command can only be run by root or the user ejabberd

All commands were executed as root.
Until version 17.04 it worked perfectly.

jarbelix:
```
grep ejabberd /etc/passwd
ejabberd:x:1001:1001::/opt/ejabberd-17.07:/bin/sh

grep ejabberd /etc/group
ejabberd:x:1001:

```

ambernetas:
Same issue on Amazon Linux 201.07 and Centos 7 OS with 17.07

jarbelix:
Worked perfectly. Thank you.

cromain:
I have simplified the user check, dropping gid matching and opted for a simpler uid only approach as described in #1716
This case allows to run either by root or install_user (if defined) and only allow current user if install_user not defined (development install for example)

PR #1858, opened 2017-07-11
fix some of the errors in the new ejabberdctl version

mathiasertl:
ejabberdctl was broken in ejabberd 17.06 or 17.07 if executed as root user.

The change in line 24 fixes the check for the user root (the if-branch of that statement (line 21/22) tests for the user $INSTALLUSER). 

The second change (line 106ff) cd's to SPOOL_DIR (otherwise ejabberdctl prints error log statements) and fixes a quotation issue (the original has a missing quote before `$@"`.

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -21,7 +21,7 @@ if [ -n "$INSTALLUSER" ] ; then
     if [ $(id -g) -eq $(id -g $INSTALLUSER || echo -1) ] ; then
         EXEC_CMD="as_current_user"
     else
-        id -Gn | grep -q wheel && EXEC_CMD="as_install_user"
+        id -un | grep -q root && EXEC_CMD="as_install_user"
     fi
 else
     EXEC_CMD="as_current_user"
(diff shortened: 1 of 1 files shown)


debalance:
I assume this would fix #1855

hamano:
Are you testing ejabberdctl from make rel?
It's more broken.
SCRIPT_DIR is removed here(bae24464d3ace4e845e5cc6d3ab7f9bf6356a026) and added here(b79aef3bbc824d43a41258758842d46fe3d2399e) and then removed here(cbe6553baa6fea9df2c8ba711247b2bb99413f2d)

cromain:
I have simplified the user check, dropping gid matching and opted for a simpler uid only approach as described in #1716

javajack:
This is still broken. When I use the DEB version available for download on Ubuntu 14.04.
From here Linux x86 64-bits DEB Package
https://www.process-one.net/downloads/downloads-action.php?file=/ejabberd/17.07/ejabberd_17.07-0_amd64.deb

https://www.process-one.net/en/ejabberd/downloads/

Issue #1184, opened 2016-07-04
Old route record in mnesia's route table haven't been remove when restarting in some cases


yzf:
> What version of ejabberd are you using?

16.04

> What operating system (version) are you using?

debian 8

> How did you install ejabberd (source, package, distribution)?

source

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

Recently, I study the route process of ejabberd, as we know, ejabberd routes messages according to the records in mnesia's table `route`. But I happened to find that, some old records in `route` were left behind after I restarted ejabberd, which means that the progresses whose pid is equal to these record's `pid` were not exist any more. Luckly, in ejabberd, these records happen to use the `local_hint` to route instread of sending messages to the `pid`'s progress, **this problem hasn't lead to any error**.

After do some tests, I finally found that this problem is related to the following thins:
1. Cluster
2. `master` node have restarted ever

In my case, I have two nodes in my cluster, `node_1` and `node_2`. And I call `ejabberdctl join_cluster 'ejabberd@node_1'` in `node_2`. If I restart the slave node `node_2` using `ejabberdctl restart`, everything is ok, the old record of `node_2` in mnesia table route is deleted. But if I restart the master node `node_1` using `ejabberdctl restart`, the old record of `node_1` in mnesia table is left behind, after that, as `node_2` become the master node and `node_1` become the slave node, this time, restarting `node_2` will case record left behind. **It seem that the master node should not restart when there are some slave nodes.**

So, after some restarting, the records in table route:

```
(ejabberd@node_1)1> mnesia:dirty_read(route, <<"xxx.xxx.xxx">>).
[{route,<<"xxx.xxx.xxx">>,<13073.278.0>,
        {apply,ejabberd_local,route}},
 {route,<<"xxx.xxx.xxx">>,
        <<"xxx.xxx.xxx">>,<0.320.0>,
        {apply,ejabberd_local,route}},
 {route,<<"xxx.xxx.xxx">>,
        <<"xxx.xxx.xxx">>,<13073.7790.150>,
        {apply,ejabberd_local,route}},
 {route,<<"xxx.xxx.xxx">>,
        <<"xxx.xxx.xxx">>,<0.16033.3>,
        {apply,ejabberd_local,route}},
 {route,<<"xxx.xxx.xxx">>,
        <<"xxx.xxx.xxx">>,<13073.30448.153>,
        {apply,ejabberd_local,route}},
[...]

PR #1871, opened 2017-07-17
Fix old route record in mnesia's route table haven't been remove when restarting in some cases (#1184)

pdorofiejczyk:
This PR fixes issue #1184. Process monitoring setup at router module startup does not work correctly now. There is mnesia query issue in init function which causes empty set as a result. This bug was created with commit https://github.com/processone/ejabberd/commit/357e48fb6b60b72f050a438164a1d5590caca8f0 where where number of columns in route table has changed.


zinid:
Thanks a lot

klomo24:
when will this fix be made available in a binary release?

PR #1888, opened 2017-07-26
Ability to filter passwords from the log in mod_http_api

madkin10:
## Things Done:
- Use `ejabberd_config:may_hide_data` function to filter passwords from the log file.

## How to Test:
- Set the hide_sensitive_log_data config option in ejabberd.yml
```
hide_sensitive_log_data: true
```
- Submit a register user request.
```
curl -H "Content-Type: application/json" -H "Authorization: AUTHTOKEN" -XPOST 127.0.0.1:5280/api/register -d '{"user": "New User", "host":"HOST", "password":"password"}'
```

- Check that the ejabberd.log filtered the password:
```
2017-07-26 14:02:55.481 [info] <0.414.0>@mod_http_api:log:542 API call register [{<<"user">>,<<"fb90c36b-06d2-4194-9273-98e064780389">>},{<<"host">>,<<"HOST">>},{<<"password">>,"hidden_by_ejabberd"}] from 127.0.0.1:60030
```

### DIFF
--- a/src/mod_http_api.erl
+++ b/src/mod_http_api.erl
@@ -538,20 +538,16 @@ json_error(HTTPCode, JSONCode, Message) ->
 
 log(Call, Args, {Addr, Port}) ->
     AddrS = misc:ip_to_list({Addr, Port}),
-    Filter = gen_mod:get_module_opt(global, ?MODULE, hide_passwords_in_logfile, false),
-    ?INFO_MSG("API call ~s ~p from ~s:~p", [Call, filter_passwords(Args, Filter), AddrS, Port]);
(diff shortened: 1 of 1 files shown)


prefiks:
So we already have option that can be used here "hide_sensitive_log_data", and there is function that does that ejabberd_config:may_hide_data(String), i would prefer if this would be used instead of adding new option for it.

madkin10:
@prefiks Didn't realize that configuration option existed. I have updated the logic to use the existing option.

prefiks:
Ok merged, thank you for your contribution.

PR #1912, opened 2017-08-04
fix FIREWALL_WINDOW option

mathiasertl:
The extending of `ERLANG_OPTS` for the `FIREWALL_WINDOW` option does not work properly. This pull request fixes the issue.


badlop:
I see something like:
> ERLANG_OPTS="...... /"

Is that final / character really desired, or is it a typo?

mathiasertl:
Oh no that was a typo. I updated the pull request!

PR #1938, opened 2017-08-13
Let 'domain_certfile' take higher precedence

weiss:
If a `domain_certfile` is specified, use that instead of the `s2s_certfile` (or `c2s_certfile`).  This is how ejabberd behaved [until][1] version 17.01.

[1]: https://github.com/processone/ejabberd/commit/309bdfbe285c82726d2ce1406fc26c19a6b37bd9


zinid:
Fine by me

PR #1944, opened 2017-08-15
Fix Docker erlang version dependency by using erlang-solutions package 1:17.5.3.

aronahl:
The Docker build was failing:
```
+ cd ejabberd
+ chmod +x ./autogen.sh
+ ./autogen.sh
+ ./configure --enable-user=ejabberd --prefix=/ --enable-all --disable-tools --disable-pam --disable-elixir
checking whether make sets $(MAKE)... yes
checking for a BSD-compatible install... /usr/bin/install -c
checking for a sed that does not truncate output... /bin/sed
checking for erl... /usr/bin/erl
checking for erlc... /usr/bin/erlc
checking for epmd... /usr/bin/epmd
checking for erl... /usr/bin/erl
checking for erlc... /usr/bin/erlc
checking Erlang/OTP version... failed
configure: error: in `/tmp/ejabberd':
configure: error: 6.2 found, 6.4 (Erlang/OTP 17.5) required
See `config.log' for more details
```

The Dockerfile already added the erlang-solutions key for apt, so I just updated sources.list and then pulled 1:17.5.3 packages (and dependencies).


cromain:
we do not maintain that Dockerfile, and will not apply this change in master.
you can still suggest such PR to rroemhild directly.
we prefer to focus on these containers https://github.com/processone/docker-ejabberd

Issue #1945, opened 2017-08-16
Ejabberd broadcasts outdated data on PubSub node configuration change


sailerinteractive:
> What version of ejabberd are you using?

17.08

> What operating system (version) are you using?

Ubuntu 14.04

> How did you install ejabberd (source, package, distribution)?

Source

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

When i enable "notify_config" for a PubSub node, Ejabberd broadcasts node configuration changes to all subscribers. However, the config included in the stanza is the old one and not the new just updated config.

PR #1950, opened 2017-08-17
Broadcast updated Pubsub node config on config change 

sailerinteractive:
As described in #1945 ejabberd is currently not sending the updated pubsub node configuration on configuration change. This i my attempt to fix it. 

Disclaimer: Erlang newbie here, but it seems to work fine so far.


cromain:
I applied a better fix on 7ad525b5, transaction function returns the new configuration to broadcast.

PR #1959, opened 2017-08-22
GSoC 2017: Support for "Let's Encrypt" ACME protocol

angelhof:
# Ejabberd support for "Let's Encrypt" ACME protocol

This is the final pull request for this: https://summerofcode.withgoogle.com/projects/#5214033996152832 GSoC 2017 project.

## Project Outline

This project implements support for the "Let's Encrypt" ACME protocol. It is a protocol that allows for certificates to be acquired in an easy, automated way, without any human intervention. So the main functionality of the implementation is acquiring and managing the aforementioned certificates. Below I will describe in detail all the functionality that this project implements.

## Detailed Description

There are two typical scenarios that the client has to support:
* Certificate Acquisition/Renewal
* Certificate Revocation

Despite being described thoroughly in the latest ACME draft: https://tools.ietf.org/html/draft-ietf-acme-acme-07 I will try to give a brief overview of those two scenarios.

In order to acquire a certificate for a domain, a client has to:
1. Create a new account at the Certificate Authority (CA) using a private key of their choice.
2. Solve a challenge, that the CA has issued, in order to verify the control over the domain.
3. Create a Certificate Signing Request (CSR) in order to request a certificate from the CA.
4. Acquire the certificate from the CA.

In order to revoke a certificate, the client has to just send a revocation request.

Knowing the above I decided to implement 4 commands that the user can use in order to acquire, manage and inspect certificates for their domains.
* get_certificate
* renew_certificate
* list_certificates
* revoke_certificate

### Get Certificate

This is the most essential command that the client uses. It acquires a certificate for all hosts specified in the configuration file or for a specific set of hosts that the user requests. This command basically follows all the steps of the first scenario above.

### Renew Certificate

The renew_certificate command renews all managed certificates that are 30 days (or closer) to expiration. This command should be scheduled for execution every day or so, in order to make sure that all certificates are up to date.

### List Certificates

This command pretty prints all managed certificates in plain or verbose mode. Specifically, in plain mode, it prints their domain, SANS, expiration date and path. In verbose it prints the domain, SANs, expiration date and certificate/key.

### Revoke Certificate

This command revokes a managed certificate for a specified domain. This command should be used when there exists serious concern that a certificate/key has been compromised.

## Structural Description

The implementation is split in 3 modules that are clearly separated from each other.
* acme_challenge.erl
* ejabberd_acme_comm.erl
* ejabberd_acme.erl

### acme_challenge

This module deals with solving the challenges that the CA issues. The only challenge type that is currently supported is "http-01". If more challenge types are to be supported they should be included in this module.

### ejabberd_acme_comm

This module contains functions that implement all necessary http requests to the ACME CA. Its purpose is to facilitate the ACME client implementation by separating the handling/validating/parsing of all the requests from the application logic.

### ejabberd_acme

This is the main module of the project. It contains implementations of all the aforementioned commands and the certificate management. It only contains application logic and calls functions from ejabberd_acme_comm in order to communicate with the CA.

## Example Usage

An example scenario would be to first execute
```
ejabberdctl get_certificate all
```

And then have a script be scheduled to run the following command every day or so
```
ejabberdctl renew_certificate
```

## Known Issues / Things left to do

The implementation offers basic support of the ACME protocol. However it is not complete and has some known issues. I will try to include all of the issues and future tasks here.

### Issues

The main issue is that I haven't tested the client on the real "Let's Encrypt" ACME CA because I don't have control of any domain. That is why I have only tested the implementation with a local CA by redirecting all domains to localhost. Because of that I am not sure whether everything works correctly on the real CA.

### Future Tasks

* Write documentation in processone/doc.ejabberd.im
* Support more key types
* Support more challenge types (dns-01, tls-sni-01)

Any feedback on this project is greatly appreciated.


Neustradamus:
A good news, it was a beautiful GSoC project!

Neustradamus:
It will be added in next version?

zinid:
Nothing works. I spent a couple of hours fixing glitches and it still doesn't work. I give up.

angelhof:
@zinid What is the exact problem? Can you explain to me so that I can help fix it?

zinid:
@angelhof for example, this call doesn't work:
```erl
> ejabberd_acme_comm:directory("https://acme-v01.api.letsencrypt.org/").
** exception error: bad argument
     in function  bitstring_to_list/1
        called as bitstring_to_list({[{<<"terms-of-service">>,
                                       <<"https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf">>}]})
     in call from ejabberd_acme_comm:'-get_dirs/1-lc$^0/1-0-'/1 (src/ejabberd_acme_comm.erl, line 153)
     in call from ejabberd_acme_comm:'-get_dirs/1-lc$^0/1-0-'/1 (src/ejabberd_acme_comm.erl, line 154)
     in call from ejabberd_acme_comm:get_dirs/1 (src/ejabberd_acme_comm.erl, line 153)
```

angelhof:
@zinid Ok I found out what the problem is. Since the time I made this pr the Let's Encrypt server has been updated and so my code does not work now as the server returns data in a different format. I will try to fix all of the problems and get back to you as soon as possible :)

zinid:
OK

zinid:
It's also unclear what should be in `contact` option.

angelhof:
@zinid As also noted in [ACME draft Section 6.1.2](https://tools.ietf.org/html/draft-ietf-acme-acme-04#section-6.1.2) contact is a URI that the server can use to contact the client in case of issues related to authorization. As it is stated in [ACME draft Section 6.3](https://tools.ietf.org/html/draft-ietf-acme-acme-04#section-6.3) the contact field should be included in a new registration.

zinid:
So we should require admins to read internet drafts to configure ejabberd?

zinid:
Why there is no default? Why there is no default for `ca_url` neither?

angelhof:
I will put the Let's Encrypt CA url as a default on `ca_url`. However I don't believe that we could have a default for `contact` as it is different for every administrator. Instead I will explain in the example configuration file what contact should be and how should it be completed.

zinid:
In my turn I don't believe an admin who is lazy to run certbot will configure this.

angelhof:
In my opinion this is not the case. The simplest way to obtain certificates using certbot is to execute a command explicitly stating all the domain names, such as this `./certbot-auto --apache -d example.com -d www.example.com -d other.example.net` (taken from [Certbot Documentation](https://certbot.eff.org/docs/intro.html#how-to-run-the-client)). In addition certbot also requires an email address when run for the first time. 

In contrast our implementation just needs writing an email address as the domain names are already written in the ejabberd configuration. So the only thing an administrator has to do is to add their email address.

weiss:
Will Let's Encrypt registration fail if we omit the contact?  If they don't enforce it, we shouldn't either, in my opionion.  (I don't remember having specified a contact address when using [acme-tiny][1], for example.)

[1]: https://github.com/diafygi/acme-tiny

angelhof:
@weiss No it will not fail, however the administrator will not be notified in any case of server-initiated certificate revocation and so on. So I guess we could get away with proposing (but not enforcing) configuring it.

zinid:
> So I guess we could get away with proposing (but not enforcing) configuring it.

Agreed.

angelhof:
I made the proposed changes. Now there is no need to configure anything to obtain certificates.

zinid:
I still get the error even with the recent commit:
```
1> ejabberd_admin:get_certificate([<<"zinid.ru">>]).
{error,configuration}
...
10:58:28.100 [error] No acme configuration has been specified
```

zinid:
When I set a contact, I get:
```
> ejabberd_admin:get_certificate([<<"zinid.ru">>]).
{error,configuration_ca_url}
...
11:01:37.484 [error] No CA url has been specified
```
WTF?

zinid:
And I definitely have the recent commit:
```
> git show --stat 78f494dd2eba0970b14247f1027f09b6a82efaf0
commit 78f494dd2eba0970b14247f1027f09b6a82efaf0
Author: Konstantinos Kallas <konstantinos.kallas@hotmail.com>
Date:   Sat Nov 11 15:38:47 2017 +0200

    Configuration file changes
    Explain the acme configuration options

 ejabberd.yml.example | 24 ++++++++++++++----------
 1 file changed, 14 insertions(+), 10 deletions(-)
```

zinid:
So I set all the options and now get:
```
1> ejabberd_admin:get_certificate([<<"zinid.ru">>]).
"Error with one or more certificates\n  Error for domain: \"zinid.ru\",  with reason: 'authorization'"
```
Well, maybe there are some problems, but it's **absolutely** unclear what to fix as the diagnostic message is cumbersome.

angelhof:
The errors are logged in a more "verbose" way in the ejabberd log, so the best place to look for an explanation would be there. 

Regarding the authorization error the CA tries to check if you have solved the challenge that was posed, so "Let's Encrypt" will try to fetch a file from zinid.ru/.well-known/ 

If your PC serves this domain, ejabberd should be able to provide the correct file to the CA (through acme_challenge.erl) so that the authorization succeeds. 

Note: Any advice on how to improve the error messages is welcome :)

zinid:
> so "Let's Encrypt" will try to fetch a file from zinid.ru/.well-known/

Looking at the code I don't see anything related to "well-known". So it's assumed an admin should configure ejabberd listener to accept request to /.well-known ?

angelhof:
At the moment I have put this in ejabberd.yml.example. Is there any better way to do this?

zinid:
I think it's better to add a `request_handler` to the existing http listener (if found). And, if not found, report the error (with some hints) and stop. We could start the listener automatically, but it's doubtful there will be any redirection from 80th port towards it.

zinid:
An admin have to configure a http listener anyway if s/he wants to provide modern IM (mod_http_upload is required in this case), so we can somewhat rely on it, IMO.

zinid:
Another possible way is to perform SNI-based TLS challenge, but I'm not sure if it's possible to provide 5222th port for this and also it requires to improve `fast_tls`.

angelhof:
So to get this straight, I will add a ```request handler``` in ejabberd_http.erl . However in order for this to work ```ejabberd_http``` should listen to port 80 or there should be some kind of redirection to it, so I also have to explain this in the example configuration.

zinid:
> So to get this straight, I will add a request handler in ejabberd_http.erl

Yes. But make sure you put it in the head of the list.

> However in order for this to work ejabberd_http should listen to port 80 or there should be some kind of redirection to it, so I also have to explain this in the example configuration.

Yes.

zinid:
Merged with some fixes.
However, pending issues still remain:
1) Interaction with ejabberd_pkix is currently borked
2) renew_certificates command should be called authomatically
3) It seems like "all" argument doesn't work for get_certificate command, but maybe I'm doing something wrong

zinid:
@angelhof I just found that intermediate certificate is not stored, why? And how to obtain it?

zinid:
Another issue when calling get_certificate() twice:
```
2017-11-17 08:28:33.600 [error] <0.732.0>@acme_challenge:ets_get_key_authorization:122 Unable to serve key authorization in: [<<"acme-challenge">>,<<"1ILvVy4wWEEPop462YLjaUdmTBZ...">>]
```

zinid:
Ah, that's because the code doesn't work with "staging" urls, great.

angelhof:
@zinid 

> I just found that intermediate certificate is not stored, why? And how to obtain it?

Should I store the intermediate certificate in the same pem file with the acquired certificates? The intermediate and root certificates can be found here [Let's Encrypt Certificates](https://letsencrypt.org/certificates/). I suppose that we need the X3 IdenTrust cross-signed.

> It seems like "all" argument doesn't work for get_certificate command, but maybe I'm doing something wrong

There are two ways to call `get_certificates all`

-  `(ejabberd@localhost)1> ejabberd_admin:get_certificate("all").`
- `$ ejabberdctl get_certificate all`

> Another issue when calling get_certificate() twice:

I have never encountered this error, I have to check it.

> Ah, that's because the code doesn't work with "staging" urls, great.

What exactly do you mean with this? Can you elaborate?

zinid:
@angelhof so intermediate certificates should be hard coded or what? Can't we get them via ACME protocol? I see in the I-D that the ACME server may return several certificates (I guess including intermediate ones).

angelhof:
@zinid There is no need to hardcode them, we could get the intermediate certifcate everytime we need it  from https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem.txt .

zinid:
@angelhof but the code becomes Let's Encrypt only, so this is not ACME anymore.

zinid:
Is it an issue of the LE ACME server implementation?

zinid:
See this: https://tools.ietf.org/html/draft-ietf-acme-acme-07#section-7.4.2

angelhof:
Ok it can be done. I will implement it. So the certificates that we save should contain the whole certificate-chain right?

zinid:
Yes, just store them inside the same file (the order doesn't matter).
Please `git pull` before doing this, also it would be great to fix this by Monday as we're releasing 17.11 this day.

angelhof:
@zinid There is a problem, when I save the end certificate, the issuer certificate, and the private key for the end certificate in a pem file and try to add it, `ejabberd_pkix:build_chain_and_check` returns the following error.

```
[error] <0.376.0>@ejabberd_pkix:build_chain_and_check:368 Failed to build certificate chain for .../ejabberd/certs/acme/test.free.pem: no matching private key found for certificate in the chain
```

zinid:
Are you sure there is a private key?

zinid:
Also, I don't have such problem, I just have the warning about unknown CA (due to missing intermediate cert).

angelhof:
This problem appeared now that I am trying to include the issuer certificate in the pem file.
The pem file looks like this

```
-----BEGIN EC PRIVATE KEY-----
....
-----END EC PRIVATE KEY-----

-----BEGIN CERTIFICATE-----
.....
-----END CERTIFICATE-----

-----BEGIN CERTIFICATE-----
.....
-----END CERTIFICATE-----
```

zinid:
Well I really cannot say what's going on without seeing these PEMs. Maybe private key is not for this certfile? What happens if you change the order?

angelhof:
I changed the order and nothing happens. If you want to have a look I could make a new pull request with the code so that you can check it out.

zinid:
Can't you just mail me this PEM file privately at ekhramtsov@process-one.net?

zinid:
I got the PEM file, I'm checking.
Meanwhile, why don't I have it working with the staging url, i.e.:
```yaml
acme:
    ca_url: "https://acme-staging.api.letsencrypt.org"
```
I have an error:
```
2017-11-17 08:28:33.600 [error] <0.732.0>@acme_challenge:ets_get_key_authorization:122 Unable to serve key authorization in: [<<"acme-challenge">>,<<"1ILvVy4wWEEPop462YLjaUdmTBZ...">>]
```

zinid:
I found a problem with private key mismatch: that's because the certfile you receive is not signed by the intermediate certfile provided, because it's signed by "CN = h2ppy h2cker fake CA", but intermediate certfile is issued for "CN = happy hacker fake CA".

zinid:
Can you please publish the patch somewhere? I will check with real domains.

angelhof:
It is this branch https://github.com/angelhof/ejabberd/tree/acme_keep_issuer_cert

zinid:
Sigh, I need now to mess with your branches?

angelhof:
I can make a pr if you prefer

zinid:
Is it so hard to publish `git diff ...`?

zinid:
No matter, I got the idea from your branch's commits.

zinid:
OK, your fixes make it work with real domains, thank you.

zinid:
Another problem:
```
> ejabberd_acme:renew_certificates().
16:17:32.697 [warning] No acme configuration has been specified
{error,get_certificates}
(ejabberd@localhost)2> 16:17:32.697 [error] No CA url has been specified in configuration
16:17:32.699 [error] Unknown error:function_clause, [{ejabberd_acme,format_get_certificate,[{ok,<<"some.domain.net">>,no_expire}],[{file,"src/ejabberd_acme.erl"},{line,220}]},{ejabberd_acme,'-format_get_certificates_result/1-lc$^1/1-0-',1,[{file,"src/ejabberd_acme.erl"},{line,207}]},{ejabberd_acme,format_get_certificates_result,1,[{file,"src/ejabberd_acme.erl"},{line,207}]},{ejabberd_acme,renew_certificates,0,[{file,"src/ejabberd_acme.erl"},{line,378}]},{erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,674}]},{shell,exprs,7,[{file,"shell.erl"},{line,687}]},{shell,eval_exprs,7,[{file,"shell.erl"},{line,642}]},{shell,eval_loop,3,[{file,"shell.erl"},{line,627}]}]
```

angelhof:
This is fixed by this:

```
diff --git a/src/ejabberd_acme.erl b/src/ejabberd_acme.erl
index cc64703..613fb84 100644
--- a/src/ejabberd_acme.erl
+++ b/src/ejabberd_acme.erl
@@ -112,9 +112,9 @@ get_commands_spec() ->
                        args_example = ["all | www.example.com;www.example1.net"],
                        args = [{domains, string}],
                        result = {certificates, string}},
-     #ejabberd_commands{name = renew_certificate, tags = [acme],
+     #ejabberd_commands{name = renew_certificates, tags = [acme],
                        desc = "Renews all certificates that are close to expiring",
-                       module = ?MODULE, function = renew_certificate,
+                       module = ?MODULE, function = renew_certificates,
                        args = [],
                        result = {certificates, string}},
      #ejabberd_commands{name = list_certificates, tags = [acme],
@@ -221,7 +221,7 @@ format_get_certificate({ok, Domain, saved}) ->
     io_lib:format("  Certificate for domain: \"~s\" acquired and saved", [Domain]);
 format_get_certificate({ok, Domain, not_found}) ->    
     io_lib:format("  Certificate for domain: \"~s\" not found, so it was not renewed", [Domain]);
-format_get_certificate({ok, Domain, exists}) ->    
+format_get_certificate({ok, Domain, no_expire}) ->    
     io_lib:format("  Certificate for domain: \"~s\" is not close to expiring", [Domain]);
 format_get_certificate({error, Domain, Reason}) ->
     io_lib:format("  Error for domain: \"~s\",  with reason: \'~s\'", [Domain, Reason]).
```

zinid:
Indeed, thank you very much!

PR #2023, opened 2017-09-30
Line endings consistency edits in the example config file

rom1dep:
This PR is split into two changesets, depending on trailing character, so if either is of any significance for parsing, it cant be easily backed-out.

This PR is purely for usability purpose, so end-users can have an easier time diffing this config-file across versions (e.g. when upgrading), especially considering the whitespace sensitivity of the file format and the fact that server-side editors may not be as well configured.

### DIFF
--- a/ejabberd.yml.example
+++ b/ejabberd.yml.example
@@ -124,7 +124,7 @@ hosts:
 ## by and what options to start it with.
 ##
 listen:
-  - 
+  -
     port: 5222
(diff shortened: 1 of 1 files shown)


rom1dep:
@zinid if you care about reviewing this one ;)

cromain:
I agree we should apply this PR.
note: our installers already use a configuration without trailing spaces.

PR #2065, opened 2017-10-24
Support for --enable-graphics in mix.exs

jorjeb:
The graphics option is set to true in vars.config when ejabberd is reconfigure with --enable-graphics, but mix.exs  currently does not check for it.


zinid:
Thanks

PR #2111, opened 2017-11-16
Add a workaround for FreeBSD sed

wahjava:
I'm the maintainer for FreeBSD port `net-im/ejabberd`, and while updating it to 17.09, and I come across a quirk about FreeBSD `sed(1)`, that it does not interpret `\|` specially, when `sed(1)` is in basic regular expression mode:

```
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |sed  -n -e '/\(\|\)/p'
Some\|thing
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |sed  -n -e '/\(M\|p\)/p'
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |sed  -n -e '/\(M\)/p'   
Makefile
```

`sed(1)` in extended-regular expression mode matches it fine though:

```
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |sed  -r -n -e '/(M|p)/p'
Makefile
patch
```

And GNU `sed(1)` works fine with both expressions: 

```
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |gsed  -n -e '/\(M\|p\)/p'
Makefile
patch
λ printf "Some\\|thing\nMakefile\npatch\nFoo\nBar\n" |gsed -r -n -e '/(M|p)/p' 
Makefile
patch
```

I will check, and file bug on FreeBSD side in case this is an unacceptable behaviour, but in the meantime, if we can commit the attached diff, that'll be great.

In case, you're not open to accepting this workaround, then in that case please close this pull-request, and I will just add this workaround downstream, which I am adding for `17.09` port anyways.

Thanks!

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -133,7 +133,7 @@ FILES_WILDCARD=$(call FILTER_DIRS,$(foreach w,$(1),$(wildcard $(w))))
 
 ifeq ($(MAKECMDGOALS),copy-files-sub)
 
-DEPS:=$(sort $(shell $(REBAR) -q list-deps|$(SED) -ne '/^[a-zA-Z0-9_-]\{1,\} \(TAG\|REV\)/s/^\([^ ]*\).*/\1/p'))
+DEPS:=$(sort $(shell $(REBAR) -q list-deps|$(SED) -rne '/^[a-zA-Z0-9_-]{1,} (TAG|REV)/s/^([^ ]*).*/\1/p'))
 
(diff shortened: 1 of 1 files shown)


cromain:
osx's sed does not recognise -r, so the change is not acceptable yet.
we are checking how this can be improved.

prefiks:
So this doesn't work on macos as sed here doesn't recognize -r option, could you please check if this will work for you?
```
sed -ne '/ TAG / s/ .*// p; / REV / s/ .*// p'
```

wahjava:
@prefiks that works, and is much simpler, and ofcourse both output identical text.

Thanks!

prefiks:
Fixed sed invocation was commited in c05626a1ba1ebae728ea90b01830582731435542

PR #2130, opened 2017-11-27
ejabberd command: Delete user MAM messages older than DAYS

oxoWrk:
removes messages from a single user message archive


zinid:
What is going on? :smile:

oxoWrk:
unskilful hands.....

zinid:
I know that feel

PR #2158, opened 2017-12-10
Use /bin/sh as the explicit shell when using su in ejabberdctl.

bowlofeggs:
Some distributions (such as Fedora) use /sbin/nologin as the login
shell for the ejabberd user. The newer version of ejabberdctl uses
su to perform the command if the INSTALLUSER invokes the script.
This commit adjusts the call to su so that it passes /bin/sh as
the shell to use so that it will work correctly when the ejabberd
user's shell is set to nologin.

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

   description.

   them as separate PRs.

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -110,7 +110,7 @@ export ERL_LIBS
 exec_cmd()
 {
     case $EXEC_CMD in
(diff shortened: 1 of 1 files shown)


cromain:
good catch ! thanks.

PR #2160, opened 2017-12-12
Support pubsub#publish-options PRECONDITIONs

weiss:
Add support for publishing options as per the [current][1] [wording][2] of XEP-0060.

Recent Conversations versions (and maybe other clients) need this to make [OMEMO][3] work without presence subscription.

[1]: https://github.com/xsf/xeps/pull/481
[2]: https://xmpp.org/extensions/xep-0060.html#publisher-publish-options
[3]: https://conversations.im/omemo/


cromain:
Thanks ! patch looks good.

PR #2167, opened 2017-12-13
Send last PEP items to owner on initial presence

### LINKED ISSUE #2112: XEP-0163 - owner did not receive any PEP notifications after initial presence stanza

17.09

CentOS Linux 7 (Core)

source

As described in [XEP-0163: Personal Eventing Protocol](https://xmpp.org/extensions/xep-0163.html) after publishing an event the owner should receive his own notifications on every resource even if the resource was offline. After the fix from yesterday [XEP-0163 - owner did not receive any PEP notifications](https://github.com/processone/ejabberd/issues/2108) (thanks for the quick turnaround) the owner only receives his notification if he is online. As described in in [XEP-163 - Notify Last](https://xmpp.org/extensions/xep-0163.html#notify-last) every new ressource of the owner should also get the notification, which is not working until now.

`As mentioned, a PEP service MUST send the last published item to all new subscribers and to all newly-available resources for each subscriber, including the account owner itself. `

Thanks

weiss:
Closes #2112.

### DIFF
--- a/src/mod_caps.erl
+++ b/src/mod_caps.erl
@@ -206,8 +206,10 @@ c2s_presence_in(C2SState,
     {Subscription, _} = ejabberd_hooks:run_fold(
 			  roster_get_jid_info, To#jid.lserver,
 			  {none, []}, [To#jid.luser, To#jid.lserver, From]),
+    ToSelf = (From#jid.luser == To#jid.luser)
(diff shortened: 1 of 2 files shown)


zinid:
I just checked the PR on my server and it indeed resolves the issue.

PR #2187, opened 2017-12-26
privacy_tests: Omit test for 'subscribe' presence

weiss:
With commit b54e1e49ba6794725bead67d530b50ab89ce078b, sending `subscribe` presence to a blocked JID generates roster pushes which the privacy tests currently fail to handle.


prefiks:
Fixed that in commit d59d4e27ec8e41c02d396881757a4129f27e3246 without disabling those tests.

weiss:
Thanks a lot! :+1:

PR #2202, opened 2018-01-03
mod_block_strangers: Bounce an error message

### LINKED ISSUE #2197: mod_block_strangers shoud bounce instead of silently dropping

17.09 (on a remote server that I'm trying to contact)

The server is deploying `mod_block_strangers` in the default configuration.

Messages sent to (existing) users on that server silently disappear. It is not clear to the sender if the messages are stored in offline storage, if the user just doesn't react, or if the messages are blocked.

Messages sent to non-existing users are bounced with `503 - cancel: Service unavailable`.

This imposes two problems:
* you don't know if your receipient got your message (which is really bad for XMPP, where we struggle with the reliability of message delivery already)
* it is possible to enumerate user accounts by differentiating between silence and 503 responses.

Please change the behavior of `mod_block_strangers` to:
* bounce blocked messages to existing users with an error
* bounce messages to non-existing accounts with the same error

A naive suggestion would be: `policy-violation` "Messages from strangers are blocked on this server."
(https://xmpp.org/rfcs/rfc6120.html#stanzas-error-conditions-policy-violation)

weiss:
Return a stanza error when a message is rejected, in order to make
legitimate users aware of the issue.

Closes #2197.

### DIFF
--- a/src/mod_block_strangers.erl
+++ b/src/mod_block_strangers.erl
@@ -79,7 +79,7 @@ filter_offline_msg({_Action, #message{} = Msg} = Acc) ->
 	deny -> {stop, {drop, Msg}}
     end.
 
-check_message(#message{from = From, to = To} = Msg) ->
+check_message(#message{from = From, to = To, lang = Lang} = Msg) ->
     LServer = To#jid.lserver,
     AllowLocalUsers =
         gen_mod:get_module_opt(LServer, ?MODULE, allow_local_users, true),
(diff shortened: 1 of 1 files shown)


zinid:
Why ?MYNAME ?

zinid:
Oops, I meant why ?MYLANG, we can fetch language tag from Msg

weiss:
> we can fetch language tag from Msg

Ah sure, I'm stupid.  Force-pushed a fix.

PR #2255, opened 2018-01-30
Add option limits to all string config fields for MUCs

kuvia:
There are already options to limit the number of characters for the MUC title and description. However, there are two more string fields that can be set by the room owner and that are not currently limitable: `password` and `captcha_whitelist`. Unlimited sizes for any field can cause Denial of Service symptoms. For example I have observed that the admin interface would not load anymore after creating a room with a huge title.

I have added two new configuration options for `mod_muc`: `max_password` and `max_captcha_whitelist`. This should be in line with the existing string limit options.

If this is fine, I will update the docs.

### DIFF
--- a/src/mod_muc.erl
+++ b/src/mod_muc.erl
@@ -1164,6 +1164,10 @@ mod_opt_type(regexp_room_id) ->
     econf:re([unicode]);
 mod_opt_type(max_room_name) ->
     econf:pos_int(infinity);
+mod_opt_type(max_password) ->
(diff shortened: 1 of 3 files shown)


himawri:
Hey, it's been a long time but I've rebased my changes onto master. Please take another look, thanks!

Neustradamus:
@kuvia: Can you rebase your PR?

prefiks:
Thanks, looks fine, thanks you for your contribution!

PR #2301, opened 2018-02-25
add ejabberd_command to get affiliation of a user in MUC room

tsaqova:
This PR propose to add functionality to get affiliation information of a user in MUC room instead of get all affiliations and expose it via ejabberd_commands.

This how it works:
- Get the room PID.
- Use mod_muc_room:get_affiliation to determine the affiliation of the user.
- Expose as ejabberd_commands.

These changes are a new capability and do not make any changes to existing code.

Here is the PR for the documentation: https://github.com/processone/docs.ejabberd.im/pull/39

### DIFF
--- a/src/mod_muc_admin.erl
+++ b/src/mod_muc_admin.erl
@@ -37,7 +37,7 @@
 	 get_user_rooms/2, get_room_occupants/2,
 	 get_room_occupants_number/2, send_direct_invitation/5,
 	 change_room_option/4, get_room_options/2,
-	 set_room_affiliation/4, get_room_affiliations/2,
+	 set_room_affiliation/4, get_room_affiliations/2, get_room_affiliation/3,
 	 web_menu_main/2, web_page_main/2, web_menu_host/3,
 	 subscribe_room/4, unsubscribe_room/2, get_subscribers/2,
 	 web_page_host/3, mod_options/1, get_commands_spec/0]).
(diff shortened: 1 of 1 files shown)


tsaqova:
@badlop 
It's ok, I just want to be more consistent with the get all affiliations API, but I cannot retrieve the `reason` field since there is no existing method to retrieve that field in the current implementation (need to refactor mod_muc_room to get it) so I add the JID and service field as a return value. What do you think?

badlop:
I don't understand your comment. Maybe I didn't write my question correctly, I'll write again.

Right now, the proposed patch includes:
```erlang
+    args_desc = ["Room name", "MUC service", "User JID"],
+    result_desc = "Affiliation of the user with username, domain, and affiliation",
+    result_example = [{"user1", "example.com", member}],
...
+    {UserJID#jid.luser, UserJID#jid.lserver, Affiliation};
```
I don't understand why the result contains username and domain, which are already known because they are provided as User JID. I propose the result to be like this:
```erlang
+    args_desc = ["Room name", "MUC service", "User JID"],
+    result_desc = "Affiliation of the user",
+    result_example = [member],
...
+    Affiliation;
```
Do you agree?

tsaqova:
I return the user JID as a result just for consistency with the existing get_room_affiliations API: https://docs.ejabberd.im/developer/ejabberd-api/admin-api#get-room-affiliations.

I've updated the PR to only return affiliation field.

badlop:
Thanks for the command and clarifications. I've merged it, and simplified a little more the format of the result.

PR #2304, opened 2018-02-25
Run SASL result callbacks earlier

weiss:
Call `Mod:handle_auth_success/4` and `Mod:handle_auth_failure/4` before sending the SASL response rather than afterwards.  This way, callbacks have a chance to send a custom response and disconnect.  (Also, it's then consistent with [the order in][1] `process_handshake/2`.)

[1]: https://github.com/processone/ejabberd/blob/22e43ebd8a82c7ffe33bbc52cb619a5e4c7def92/src/xmpp_stream_in.erl#L677


zinid:
Yes, nice idea, thanks

PR #2327, opened 2018-03-19
 Fix offline message with sql type, user alway didn't received offline messages

cisiqo:
If the record doesn't add the expire attribute, the offline message expire time always the current time, so user alway didn't received the offline message when set offline module with sql type.


badlop:
Umm... First testing: I configure mod_offline with db_type: sql, then send messages to an offline use, and they are delivered to the user when he logins. So, the feature works correctly.

And second, I don't understand what's the relation between message expiration and message delivery, and why would it affect message delivery to any expiration date that an offline message may or may not have.

PR #2351, opened 2018-03-28
config: move section about direct-tls for c2s just under regular c2s config

rom1dep:
That should ease parameters comparison between direct and traditional c2s config.


rom1dep:
Thanks!

PR #2358, opened 2018-03-30
feat: add hook for client state activity

Pouriya-Jahanbakhsh:
Hi. New hook `csi_activity` added. Callback function must accept user jid as first argument and activity (`active` and `inactive`) as second argument. Hook will run in global mode.

### DIFF
--- a/src/mod_client_state.erl
+++ b/src/mod_client_state.erl
@@ -194,10 +194,12 @@ c2s_stream_started(State, _) ->
     init_csi_state(State).
 
 -spec c2s_authenticated_packet(c2s_state(), xmpp_element()) -> c2s_state().
-c2s_authenticated_packet(C2SState, #csi{type = active}) ->
+c2s_authenticated_packet(#{jid := Jid}=C2SState, #csi{type = active}) ->
+    ejabberd_hooks:run(csi_activity, [Jid, active]),
     C2SState1 = C2SState#{csi_state => active},
     flush_queue(C2SState1);
-c2s_authenticated_packet(C2SState, #csi{type = inactive}) ->
+c2s_authenticated_packet(#{jid := Jid}=C2SState, #csi{type = inactive}) ->
+    ejabberd_hooks:run(csi_activity, [Jid, inactive]),
     C2SState#{csi_state => inactive};
 c2s_authenticated_packet(C2SState, _) ->
     C2SState.


zinid:
And this case the hook should be local :)
Also, the hook should be of type `run_fold` and should possess `State` as an accumulator, so, it should look like:
```erl
State1 = ejabberd_hooks:run_fold(csi_activity, LServer, State, [active])
```

Pouriya-Jahanbakhsh:
:rofl: OMG. But can you explain why? Is this flow for all hooks of c2s?  
Are you agree with:  
```erlang
-spec c2s_authenticated_packet(c2s_state(), xmpp_element()) -> c2s_state().
c2s_authenticated_packet(#{lserver := LServer, jid := Jid}=C2SState0, #csi{type = Type}) ->
  when Type == active orelse Type == inactive ->
	C2SState2 =
        % This pattern matching prevents saving erroneous data for 'csi_state'
        case ejabberd_hooks:run_fold(csi_activity, LServer, C2SState0, [Type]) of
            {C2SState1, active} ->
                C2SState1#{csi_state => active};
            {C2SState1, inactive} ->
                C2SState1#{csi_state => inactive};
        end,
    flush_queue(C2SState2);
c2s_authenticated_packet(C2SState, _) ->
    C2SState.
```

Pouriya-Jahanbakhsh:
You could at least say thanks and explain why merged someone else's pr. This behavior stops me for next contributions.

weiss:
> You could at least say thanks and explain why merged someone else's pr.

I'm sorry, my assumption was that you're interested in the actual functionality, and in this case it seemed quicker to just implement it than to explain the required changes.  Some explanations:

- Letting callback functions access the c2s state makes the hook more flexible.  You'll find various examples of callbacks touching the c2s state in the source code.
- The c2s state update (and the queue flushing) performed by `mod_client_state` itself is now done in a [`csi_activity` callback][1], which allows other modules to prevent (or influence) this.
- You shouldn't call `flush_queue/1` when the client went `inactive`.
- You shouldn't try to handle buggy callback functions.

Hope this helps, and thank you for your contribution.

[1]: https://github.com/processone/ejabberd/blob/5f1191b9f5aaa27141bf7a38f33b68461b39d3a8/src/mod_client_state.erl#L208

Pouriya-Jahanbakhsh:
Hmmmm. Thanks for detailed explanation.

PR #2369, opened 2018-04-07
Check packet in callback functions of 'user_send_packet' hook

Pouriya-Jahanbakhsh:
Hi. We wrote some filtering modules for c2s packets and we used `user_send_packet` hook. Our hook callback has priority 1 and sometimes it yields `drop` instead of `NewPkt`.When our module drops some packets we see lots of error messages in logs. For example [here](https://github.com/processone/ejabberd/blob/ea87bdfbe57027b2160bb4ff9d5500ae54126123/src/mod_carboncopy.erl#L143) `xmpp:get_from(drop)` causes function clause error. So i have created this PR. I preferred guard `Pkt =/= drop` over [these](https://github.com/processone/xmpp/blob/master/include/xmpp.hrl#L32) 4 guards.

### DIFF
--- a/src/mod_carboncopy.erl
+++ b/src/mod_carboncopy.erl
@@ -139,13 +139,15 @@ iq_handler(#iq{type = get, lang = Lang} = IQ)->
 
 -spec user_send_packet({stanza(), ejabberd_c2s:state()})
       -> {stanza(), ejabberd_c2s:state()} | {stop, {stanza(), ejabberd_c2s:state()}}.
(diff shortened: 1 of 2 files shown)


Pouriya-Jahanbakhsh:
All [tests](https://travis-ci.org/Pouriya-Jahanbakhsh/ejabberd) have been passed for my fork.

zinid:
Just return `{stop, drop}` from your callback. It's anyway pointless to continue hook execution when you return `drop`.

PR #2399, opened 2018-04-24
ejabberdctl: fix parameters parsing

4z3:
This PR proposes a fix for the `ejabberdctl` parameter parser, which currently does not allow passing multiple parameters of the form `--param arg`.

Example:
```console
$ ejabberdctl --spool /dir1 --logs /dir2 foreground
```

The above command line will be parsed as follows
1. `for` sets `arg` to `--spool`
1. `case $arg` sets `SPOOL_DIR` to `/dir1` and `shift`s
1. `shift` is called again; `$@` is now `--logs`, `/dir2`, and `foreground`
1. `for` sets `arg` to `/dir1` (because `shift` had no effect on the list passed to `for`)
1. `case $arg` calls `break`

The `main` part of `ejabberdctl` in turn fails because `$@` is as seen in (3).

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -41,19 +41,17 @@ case $(id -un) in
 esac
 
 # parse command line parameters
-for arg; do
-    case $arg in
-        -n|--node) ERLANG_NODE_ARG=$2; shift;;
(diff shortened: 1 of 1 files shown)


cromain:
+1
i'm not sure i even used this once... patch is good

Issue #2215, opened 2018-01-10
Route selection/balancing algorithm


pdorofiejczyk:
> What version of ejabberd are you using?

17.08

> What operating system (version) are you using?

CentOS

> How did you install ejabberd (source, package, distribution)?

package

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

Hello,

We have cluster of ejabberd behind load balancer and a few nodes of custom transport components (handling legacy solution sync). We have set domain balancing to bare_source so each stanza from given user should be balanced to the same transport component which is critical to us because we multiplicate one session in legacy solution to multiple in ejabberd. This is true when new user session is balanced to the same ejabberd node but from other nodes ejabberd selects other transport component.

As I see in ```ejabberd_router.erl``` there are two route selection algorithms. Algorithms can be switched by setting ```domain_balancing_component_number``` in config. When ```domain_balancing_component_number``` is set, only given number of nodes can be registered and routes are selected from all routes available (local and remote) but in case of transport component crash, users are not rebalanced to other nodes. In the other case  the algorithm selects routes by phash of selected domain balancing but with priority of local routes so if any local routes are available one of them would be selected.

I am thinking of making PR with implementation of consistent hasing algorithm to solve my problem and improve reliability of ejabberd routing. The question is why this kind of algorithm is not implemented yet? Do yo see any downsides of this?

zinid:
The problem with consistent hashing is that you need to do rebalancing/data migration in the case when a new node (component) arrives. E.g., when a new component connects, it's injected into hashing ring instantly and 1/N part of users will be routed to this new instance. If a client session maintains "connection state" with a particular component, this may lead to stale sessions on the component or something like that.

pdorofiejczyk:
I agree with your arguments but each algorithm has pros and cons. Let's see:

**"Static" algorithm (```domain_balancing_component_number```)**

pros:
- users are balanced always to the same component independently of ejabberd node in cluster
cons:
- in case of one component fail/disconnect, 1/N users will be not rebalanced and don't have access to component's functionality
- components can't be added dynamically to handle bigger load

**"Dynamic" algorithm with local routes priority**

pros:
- users will be rebalanced if one component fail/disconnect
- components can be dynamically added
cons:
- there is no guarantee to user will be balanced to the same component independently of ejabberd node in cluster
- nearly all users will be rebalanced if one component is added/removed

**Consistent hashing**
pros:
[...]

PR #2401, opened 2018-04-26
Route balancing using jumping consistent hashing algorithm (#2215)

pdorofiejczyk:
1. Added new balancing algorithm (detailed explanation in #2215)
2. Added configuration option  ```domain_balancing_algorithm``` with possible values: ```dynamic```, ```fix_number``` and ```consistent_hashing``` where:
- ```dynamic``` behaves like in situation when ```domain_balancing_component_number ``` has not been set
- ```fix_number``` behaves like in situation when ```domain_balancing_component_number ``` has been set
- ```consistent_hashing``` is the new algorithm explained in #2215

3. When ```domain_balancing_algorithm``` is not set, fallback to ```domain_balancing_component_number ``` logic is used

### DIFF
--- a/include/ejabberd_router.hrl
+++ b/include/ejabberd_router.hrl
@@ -2,6 +2,8 @@
 
 -type local_hint() :: integer() | {apply, atom(), atom()}.
 
+-type balancing_algorithm() :: dynamic | fix_number | consistent_hashing.
+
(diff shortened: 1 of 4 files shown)


Neustradamus:
@pdorofiejczyk: Can you rebase your PR, it will be nice to see this improvement merged...

PR #2416, opened 2018-05-14
Fix compiling issue in elixir umbrella app

Schultzer:
Fixes the issue when compiling ejabberd in an Elixir umbrella app structure where deps is located at `../../deps` instead of `deps` I do a simple check to verify that it's an umbrella app with `Mix.Project.umbrella?/0` and then return the right path. this issue is occurring when require ejabberd in `my_umbrella_app/apps/my_umbrella_app_web/mix.exs`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -82,8 +82,8 @@ defmodule Ejabberd.Mixfile do
 
   defp deps_include(deps) do
     base = case Mix.Project.deps_paths()[:ejabberd] do
-      nil -> "deps"
-      _ -> ".."
+      nil -> unless Mix.Project.umbrella?(), do: "deps", else: "../../deps"
+      _   -> unless Mix.Project.umbrella?(), do: "..",   else: "../../"
     end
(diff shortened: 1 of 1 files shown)


drewpotter:
Thanks for this work @Schultzer ! Do you know how much RAM roughly you need for compiling your Elixir umbrella app with ejabberd? Mine is reaching around 22GB RAM usage whilst `Compiling esip`.

I really hope this work can get merged into master soon and we can use a precompiled module in Elixir!

drewpotter:
I managed to get this working without using a fork! 

At the top directory of the umbrella app I put this in the mix.exs file:
```
  defp deps do
    [{:ejabberd, "~> 18.6"}]
  end
```

Then in a sub directory, under apps, I put this inside the xmpp app folder's mix.exs file:
```
  def application do
    [
      extra_applications: [:logger, :ejabberd],
      mod: {Dinarly.XMPP.Application, []}
    ]
  end
```

It fires up as expected with no errors.

Freyskeyd:
Any idea if this PR is still interesting?

PR #2446, opened 2018-05-30
Fix authentication for usernames containing uppercase characters

### LINKED ISSUE #2280: Upgrade from 16.06 to 17.08: usernames with capital letters are rejected

17.08

Ubuntu 14.04.5 LTS, Trusty Tahr

Package

We use extauth_program to authenticate users for the IM. On 16.08 usernames containing capital letters were authenticated without problems. The auth string that was sent to the external program by ejabberd was:

auth:vlasisTest:im.myhost.com:password

However, with 17.08 the extauth_program is not even called when the username contains capital letters. In such cases we only see in the ejabberd logs:

2018-02-14 07:07:11.156 [info] <0.740.0>@ejabberd_c2s:handle_auth_failure:446 (http_bind|ejabberd_bosh) Failed c2s PLAIN authentication for vlasisTest@im.myhost.com from 87.202.179.100: Invalid username or password

On the other hand, if I use an XMPP client to authenticate as vlasistest on 17.08, the extauth_program is called as expected and the authentication string that is sent is:

auth:vlasistest:im.myhost.com:password

nosnilmot:
Fix authentication for usernames containing uppercase characters

Applies to authentication methods that compare User (normalized)
and AuthzId (was not being normalized). These are external, ldap & pam.

Fixes #2280

### DIFF
--- a/src/ejabberd_auth.erl
+++ b/src/ejabberd_auth.erl
@@ -230,19 +230,22 @@ check_password_with_authmodule(User, AuthzId, Server, Password) ->
 check_password_with_authmodule(User, AuthzId, Server, Password, Digest, DigestGen) ->
     case validate_credentials(User, Server) of
 	{ok, LUser, LServer} ->
-	    lists:foldl(
-	      fun(Mod, false) ->
-		      case db_check_password(
-			     LUser, AuthzId, LServer, Password,
+	    case jid:nodeprep(AuthzId) of
+		error ->
+		    false;
+	    LAuthzId ->
+		lists:foldl(
+		  fun(Mod, false) ->
+			case db_check_password(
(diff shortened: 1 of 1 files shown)


zinid:
What if `jid:nodeprep(Authzid)` returns `error`? In this case we will have a fancy traceback in the log. I would rather suggest to perform nodeprep here: https://github.com/processone/ejabberd/blob/master/src/ejabberd_auth.erl#L233 - right after we got success from `validate_credentials` we should create another `case` branch where we call `jid:nodeprep(Authzid)`. This branch should return `false` if nodprep returns `error`.

nosnilmot:
I believe I have now implemented what you suggested

PR #2496, opened 2018-06-30
fix documentation of ejabberd.yml configuration

genofire:


badlop:
Since https://github.com/processone/ejabberd/commit/77163c43d243daadcc351a8b26ed3b285d379d80 , that file doesn't contain those lines, so this PR is no longer necessary.

cromain:
OK so let's close this PR then

Issue #1905, opened 2017-08-02
Ejabberd fails to start after issuing a call to redis : [<<"HKEYS">>,<<"ejabberd:sm:[Host]">>]


satish-olx:
> What version of ejabberd are you using?

17.07

> What operating system (version) are you using?

Ubuntu:14.04

> How did you install ejabberd (source, package, distribution)?

Package

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

Ejabberd fails to start. It generates an error log

```
=erl_crash_dump:0.3
Wed Aug  2 11:21:49 2017
Slogan: eheap_alloc: Cannot allocate 762886488 bytes of memory (of type "heap").
System version: Erlang/OTP 19 [erts-8.3.2] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:true]
Compiled: Fri May 12 17:11:05 2017
Taints: ets_cache,asn1rt_nif,jid,stringprep,fxml_stream,fxml,fast_tls,p1_sha,fast_yaml,crypto,erl_tracer
Atoms: 27515
Calling Thread: scheduler:8
=scheduler:1
Scheduler Sleep Info Flags:
Scheduler Sleep Info Aux Work: DD | DD_THR_PRGR | THR_PRGR_LATER_OP
Current Port:
Run Queue Max Length: 0
Run Queue High Length: 0
Run Queue Normal Length: 2
Run Queue Low Length: 0
Run Queue Port Length: 0
Run Queue Flags: NONEMPTY_NORMAL | NONEMPTY | UNKNOWN(134217728)
Current Process: <0.1095.0>
Current Process State: Garbing
Current Process Internal State: ACT_PRIO_HIGH | USR_PRIO_HIGH | PRQ_PRIO_HIGH | ACTIVE | RUNNING | GC | ON_HEAP_MSGQ
Current Process Program counter: 0x00007f4c543fb028 (io_lib:write_binary_body/2 + 368)
Current Process CP: 0x00007f4c543fb090 (io_lib:write_binary_body/2 + 472)
Current Process Limited Stack Trace:
0x00007f4a5e069588:SReturn addr 0x543FB090 (io_lib:write_binary_body/2 + 472)
0x00007f4a5e069598:SReturn addr 0x543FB090 (io_lib:write_binary_body/2 + 472)
0x00007f4a5e0695a8:SReturn addr 0x543FB090 (io_lib:write_binary_body/2 + 472)
[...]

PR #2525, opened 2018-07-12
Use lua scripting to clear redis sessions.

satish-olx:
I had come across https://github.com/processone/ejabberd/issues/1905. @zinid suggested to submit a PR for improving the behavior using lua.  

In the proposed new session management scheme, an additional key which associates session to node mapping is maintained in redis. This is done because we need to find out the node to session association without it being deserialized in ejabberd. A recursive call is made to redis server for cleaning all the sessions, since lua-timeout might occur in the presence of large amount of sessions.

### DIFF
--- a/src/ejabberd_sm_redis.erl
+++ b/src/ejabberd_sm_redis.erl
@@ -27,16 +27,17 @@
 -define(GEN_SERVER, p1_server).
 -endif.
 -behaviour(?GEN_SERVER).
-
+-define(DELETION_CURSOR_TIMEOUT_SEC, "30").
 -behaviour(ejabberd_sm).
 
 -export([init/0, set_session/1, delete_session/1,
 	 get_sessions/0, get_sessions/1, get_sessions/2,
-	 cache_nodes/1]).
+	 cache_nodes/1, clean_table/1, clean_table/0]).
 %% gen_server callbacks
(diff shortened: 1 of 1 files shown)


zinid:
Thanks a lot, I cleaned up the code a bit (see 420e05fa0d5a159194ce3455d4ed2aa4f2256591).
The only problem is that `redis.replicate_commands()` seems to be a relatively new function: I had to upgrade my Redis instance to make it work. Do you know if we really need to call this function?

satish-olx:
You are correct, this has been recently added in redis source for smarter
master/slave replication mechanism. In our case I tried to run it without
using the replicate command flag, but redis threw an error, which suggested
to use this. In my opinion this is because we are using HSCAN command which
might yield different results in different redis nodes, so to ensure
consistency in among redis nodes, we have to use it.

Thank You,

Regards,
Satish Chandra

zinid:
Well, this is unacceptable. From the administrator perspective it will be a regression: everything has been working smoothly before the upgrade. I don't understand why we require it on old Redis versions. Also, not everyone is running Redis with master/slave replication.

satish-olx:
We need to have this command even on a standalone redis server in the newer
redis versions. I think either redis could be upgraded or we could do a
redis server version check during runtime and run or not run this command
based on that.

PR #2526, opened 2018-07-13
Enforce pubsub option required/rejected attributes

nosnilmot:
XEP-0060 states that 'node' and 'jid' attributes to <options> element MUST NOT
be included when <options> are specified at same time as <subscribe> :

https://xmpp.org/extensions/xep-0060.html#subscriber-configure-subandconfig

mod_pubsub will require 'node' and 'jid' attributes on standalone pubsub
options requests, and reject subscribe requests that have options that include
either 'node' or 'jid'

Note: this depends on PR https://github.com/processone/xmpp/pull/34


zinid:
Also please bump `xmpp` version in `rebar.config`. I already merged that PR.

nosnilmot:
I think I have improved the error messages, and also bumped xmpp version in rebar.config. Unfortunately XEP-0060 does not define jid-forbidden and nodeid-forbidden errors, so those are just textual annotations.

Issue #2529, opened 2018-07-15
Wrong behavior of privacy lists (xep-0016)


crosser:
> What version of ejabberd are you using?
git HEAD

> What operating system (version) are you using?
ubuntu bionic

> How did you install ejabberd (source, package, distribution)?
git source or from package

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?

I asked a [quesion on serverfault](https://serverfault.com/q/921998/195042) first, but now it looks like a bug in the code:
To fend off xmpp spam, I use xep-0016 privacy list, that allows messages from contacts with mutual subscriptions (plus from muc servers), and denies by default:

```
<!-- Outgoing Sun 15 Jul 2018 14:19:00 CEST -->
<iq xmlns="jabber:client" type="get" id="1ebf80d8-0997-479a-803f-46eb5e635001" from="egc@average.org/gajim.1ZZTNLAD">
<query xmlns="jabber:iq:privacy" />
</iq>

<!-- Outgoing Sun 15 Jul 2018 14:19:00 CEST -->
<r xmlns="urn:xmpp:sm:3" />

<!-- Incoming Sun 15 Jul 2018 14:19:00 CEST -->
<iq xml:lang='en' to='egc@average.org/gajim.1ZZTNLAD' from='egc@average.org' type='result' id='1ebf80d8-0997-479a-803f-46eb5e635001'>
<query xmlns='jabber:iq:privacy'>
<active name='Antispam'/>
<default name='Antispam'/>
<list name='Antispam'/>
</query>
</iq>

<!-- Incoming Sun 15 Jul 2018 14:19:00 CEST -->
<r xmlns='urn:xmpp:sm:3'/>

<!-- Outgoing Sun 15 Jul 2018 14:19:00 CEST -->
<a xmlns="urn:xmpp:sm:3" h="127" />

<!-- Incoming Sun 15 Jul 2018 14:19:00 CEST -->
<a h='106' xmlns='urn:xmpp:sm:3'/>

<!-- Outgoing Sun 15 Jul 2018 14:19:50 CEST -->
<iq xmlns="jabber:client" type="get" id="37a7c289-1e95-4531-b892-7c46d8bf6f07" from="egc@average.org/gajim.1ZZTNLAD">
<query xmlns="jabber:iq:privacy">
<list name="Antispam" />
</query>
</iq>

<!-- Outgoing Sun 15 Jul 2018 14:19:50 CEST -->
[...]

PR #2530, opened 2018-07-15
mod_privacy: honor the 'order' attribute

crosser:
When checking a packet against the active privacy list, make sure that
the items in the list are sorted according to the 'order' attribute.

Bug-url: https://github.com/processone/ejabberd/issues/2529
Signed-off-by: Eugene Crosser <crosser@average.org>

   description.

   them as separate PRs.


zinid:
No, this is not efficient. The list should be sorted during iq-set.

crosser:
Then you would need to guarantee that the backend preserves the order. Is there such guarantee, including possible future backends?

In any case, my concern is the bug #2529 . If it is fixed in some other way, I'll be perfectly happy.

crosser:
@zinid , I pushed a new version of the patch that does not use `lists:sort()`. (But the code is more difficult to read, unfortunately :( ).  Please see if this is more acceptable.  Thanks.

Neustradamus:
@prefiks: What do you think about this PR and Issue?

PR #2538, opened 2018-07-18
Enable TLS by default (and require it for c2s)

weiss:


Neustradamus:
Why not 5281 port?

weiss:
What would be the advantage?  The mnemonic for 5443 is that 443 of course.

PR #2551, opened 2018-07-30
Make it possible to join extauth pools

MarcelWaldvogel:
When several domains/hosts are served by a single ejabberd instance, each of them starts its own extauth pool. This can result in a huge number of child processes, making reloads/restarts slow and memory-consuming.

With this PR, multiple hosts can share the extauth pool, when the administrator believes this is necessary/possible for their setup.


zinid:
Very nice idea, thanks a lot.

PR #2622, opened 2018-10-01
Rename two presence functions

aquarhead:
`send_self_presence` was actually implementing XEP-0115, [specifically this section](https://xmpp.org/extensions/xep-0115.html#advertise)

while `send_initial_presence` was instead implementing [the "self-presence" step as described in XEP-0045](https://xmpp.org/extensions/xep-0045.html#order)

I spent quite some time reading through code and several XEPs to find out what they actually do... especially the first name confused me a lot when reading through XEP-0045, I think renaming these following the narrative in respective XEPs could help avoiding these confusions in the future.

also in the [original commit](https://github.com/processone/ejabberd/commit/ffe02c46e44b52d87425a8f4dfd42b58c5a426d6#diff-9e8ad34ba1ce71132f4de59d3e228997R2069) which added `send_self_presence` it mentioned XEP-0115

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -2149,15 +2149,15 @@ presence_broadcast_allowed(JID, StateData) ->
 -spec send_initial_presences_and_messages(
 	jid(), binary(), presence(), state(), state()) -> ok.
 send_initial_presences_and_messages(From, Nick, Presence, NewState, OldState) ->
-    send_self_presence(From, NewState),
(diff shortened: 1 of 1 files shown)


mremond:
@alexeyshch @zinid What do you think of this proposal ? It seems to make sense to me.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

aquarhead:
Rebased onto `master`

PR #2634, opened 2018-10-11
Update mod_muc_room.erl

ltAldoRaine:
if user sends decline message service should set his/her affiliation to none

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -264,8 +264,7 @@ normal_state({route, <<"">>,
 	    ejabberd_router:route_error(Packet, Err),
 	    {next_state, normal_state, StateData};
 	false when Type /= error ->
-	    handle_roommessage_from_nonparticipant(Packet, StateData, From),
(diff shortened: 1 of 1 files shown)


badlop:
Hi, nice finding! Looking at your code and the XEP, I noticed another problem, and committed a fix here: https://github.com/processone/ejabberd/commit/b010a1a0a03fb304ad018b4d42d69df93ca4c5a1

Regarding your commit, I think it introduces a new problem: your code allows a member to revoke membership of another member (and this is only allowed to admins and owners). How? Simply inviting a member that hasn't joined the room: if that user declines the invitation, the room will remove its membership! Can you test to see if that problem exists?

ltAldoRaine:
@badlop I have done it intentionally. I wanted that every member in members-only room could invite another user. member could revoke membership of only invited member. If second member declines invitation it will revoke it's membership and i think it not problem.

Neustradamus:
@ltAldoRaine: Can you sign the CLA?

Issue #2186, opened 2017-12-26
Failed inbound s2s EXTERNAL authentication


97zgxgw:
> What version of ejabberd are you using?
17.11

> What operating system (version) are you using?
Ubuntu 16.04

> How did you install ejabberd (source, package, distribution)?
from ejabberd-17.11-linux-x86_64-installer.run

> What did not work as expected? Are there error messages in the log? What
> was the unexpected behavior? What was the expected result?
I create SSL cert from gandi.net, create a user and login to jabber and all fine.
I add new contact to jabber from other server and when i try authorize him in log i see this 
<0.486.0>@ejabberd_s2s_in:handle_auth_failure:206 (tls|<0.485.0>) Failed inbound s2s EXTERNAL authentication otherserver.im -> myserver.com (::FFFF:144.7.1.2): unable to get local issuer certificate

And all my contact  in offline.

weiss:
> Failed inbound s2s EXTERNAL authentication otherserver.im -> myserver.com

So you don't trust the certificate offered by `otherserver.im`.

Do you have `mod_s2s_dialback: {}` in your list of `modules`?

97zgxgw:
Yes   mod_s2s_dialback: {} enabled in modules section of my ejabberd.yml

97zgxgw:
I add 
`s2s_cafile: "/etc/ssl/certs/ca-certificates.crt"`
and my issue resolved, but i thing it is workaround ...

zinid:
A workaround to what? How to tame openssl on every platform?

zinid:
Ah, this is the installer.
The installer will be fixed by release of 17.12.

97zgxgw:
Why i need to do when 17.12 is released ?

97zgxgw:
What wrong with 17.11 installer ? I have another way to fix my issue ?

zinid:
You need nothing to do. The installer will be shipped with CA bundle (from Mozilla), and `ca_file` would be configured in the default config pointing to the bundle.

zinid:
No, `s2s_cafile` is normal workaround to your problem.

zinid:
Our installers have already Mozilla's CA bundle included and configured in `ca_file` option (and it's used as a fallback if `s2s_cafile` is not set).
I don't think it's a great idea to set the option in `ejabberd.yml.example` because the location of the CA bundle is OS specific. It's supposed that you know what you do when building ejabberd from source and you're prepared to configure more.

[...]

PR #2642, opened 2018-10-23
Add (commented) reference to ca_file in the example config

laszlovl:
As discussed in #2186: make sure that the `ca_file` option is documented in the example config.

On at least Ubuntu 18.04 systems, s2s connections that require SSL will fail if this option is not set.

I don't understand the root cause, perhaps @zinid can fill in some details. I read https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/ and checked `openssl version -a` on my system, which reports `OPENSSLDIR: "/usr/lib/ssl"`. Looks like that's valid; `/usr/lib/ssl/certs` is a symlink to `/etc/ssl/certs`..

### DIFF
--- a/ejabberd.yml.example
+++ b/ejabberd.yml.example
@@ -39,6 +39,10 @@ certfiles:
   - "/etc/letsencrypt/live/localhost/fullchain.pem"
   - "/etc/letsencrypt/live/localhost/privkey.pem"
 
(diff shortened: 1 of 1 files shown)


zinid:
After several discussions the agreement was not to include any commented options into the example configuration, so I cannot merge this PR, sorry.

PR #2651, opened 2018-10-27
Fix mod_privacy race condition

nbruning:
mod_privacy updates the c2s state in user_receive_packet, which
tracks the *result* of the IQ set for active privacy lists.

When a second stanza is sent directly after a privacy list request,
the second stanza will be processed using the old privacy list,
because the IQ result has not yet been routed.

To reproduce, sent this XML in one go:

```xml
<iq xmlns="jabber:client" type="set" id="id1">
   <query xmlns="jabber:iq:privacy">
      <list name="invisible">
         <item action="deny" order="1">
            <presence-out />
         </item>
      </list>
   </query>
</iq>
<iq xmlns="jabber:client" type="set" id="id2">
   <query xmlns="jabber:iq:privacy">
      <active name="invisible" />
   </query>
</iq>
<presence />
```

Observe that ejabberd will broadcast presence for the user, but shouldn't.

In Ejabberd 16.x, this issue was not present. It's probably introduces by 35d19b32f40a4f6aedecca36c3145ab3013ecf87 or 1e55e018e.

Furthermore, this post seems to imply that ejabberd should *fully* handle the iq set before processing the presence packet: https://www.ietf.org/mail-archive/web/xmpp/current/msg01781.html

Front logo [Front conversations](https://app.frontapp.com/open/top_17gxw)

### DIFF
--- a/src/mod_privacy.erl
+++ b/src/mod_privacy.erl
@@ -434,7 +434,7 @@ user_send_packet({#iq{type = Type,
             % Adjust the client's state directly, so the next to-be-processed
             % packet will take the active list into account.
             {IQ, State#{privacy_active_list => Active}};
-          true ->
+          _ ->
             {IQ, State}
         end;
       _ -> {IQ, State}
--- a/src/mod_privacy.erl
+++ b/src/mod_privacy.erl
@@ -419,6 +419,27 @@ user_send_packet({#iq{type = Type,
 		false -> IQ
 	    end,
     {NewIQ, State};
+
+user_send_packet({#iq{type = Type,
+          from = #jid{luser = U, lserver = S},
+          to = #jid{luser = U, lserver = S},
+          sub_els = [_]} = IQ,
+        State})
+  when Type == set ->
+    case xmpp:get_subtag(IQ, #privacy_query{}) of
+      #privacy_query{active = undefined} -> {IQ, State};
+      #privacy_query{default = undefined, active = Active} ->
+        case get_user_list(U, S, Active) of
+          {ok, _} ->
+            % Adjust the client's state directly, so the next to-be-processed
+            % packet will take the active list into account.
+            {IQ, State#{privacy_active_list => Active}};
+          true ->
+            {IQ, State}
+        end;
+      _ -> {IQ, State}
+    end;
+
 user_send_packet(Acc) ->
     Acc.
 
--- a/src/mod_privacy.erl
+++ b/src/mod_privacy.erl
@@ -36,7 +36,7 @@
 	 check_packet/4, remove_user/2, encode_list_item/1,
          get_user_lists/2, get_user_list/3,
 	 set_list/1, set_list/4, set_default_list/3,
-	 user_send_packet/1, user_receive_packet/1,
+	 user_send_packet/1,
 	 import_start/2, import_stop/2, import/5, import_info/0,
 	 mod_opt_type/1, mod_options/1, depends/2]).
 
@@ -78,8 +78,6 @@ start(Host, Opts) ->
 		       c2s_copy_session, 50),
     ejabberd_hooks:add(user_send_packet, Host, ?MODULE,
 		       user_send_packet, 50),
-    ejabberd_hooks:add(user_receive_packet, Host, ?MODULE,
-		       user_receive_packet, 50),
     ejabberd_hooks:add(privacy_check_packet, Host, ?MODULE,
 		       check_packet, 50),
     ejabberd_hooks:add(remove_user, Host, ?MODULE,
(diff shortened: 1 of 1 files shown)


badlop:
Did you test your patch with recent ejabberd git? It crashes just at client login:

> 11:40:41.337 [error] Hook user_send_packet crashed when running mod_privacy:user_send_packet/1:
> ** Reason = {error,{case_clause,false},[
> {mod_privacy,user_send_packet,[{file,"src/mod_privacy.erl"},{line,429}],1},
> {ejabberd_hooks,safe_apply,[{file,"src/ejabberd_hooks.erl"},{line,380}],4},
> ...

nbruning:
@badlop sorry, there was a little Erlang mistake in the patch.
Now it correctly ignores non-privacy list IQ's.

badlop:
I can only see the old commit from two weeks ago, I don't see anyone recent. maybe you forgot to push the changes? Otherwise I can't find it...

nbruning:
I updated the commit itself (amend), apparently github still displays the original commit date.
Commit 3478d2296f218578f0fccfa791e620133e050b25 is actually the updated, non-crashing version.

badlop:
Umm, I see this commit, last updated 7 november, which still makes tkabber crash at login:

```diff
From 3478d2296f218578f0fccfa791e620133e050b25 Mon Sep 17 00:00:00 2001
From: Nathan Bruning <nathan@iperity.com>
Date: Sat, 27 Oct 2018 15:26:37 +0200
Subject: [PATCH] Fix mod_privacy race condition

mod_privacy updates the c2s state in user_receive_packet, which
tracks the *result* of the IQ set for active privacy lists.

When a second stanza is sent directly after a privacy list request,
the second stanza will be processed using the old privacy list,
because the IQ result has not yet been routed.
---
 src/mod_privacy.erl | 21 +++++++++++++++++++++
 1 file changed, 21 insertions(+)

diff --git a/src/mod_privacy.erl b/src/mod_privacy.erl
index 6cb57343..19c74fe8 100644
--- a/src/mod_privacy.erl
+++ b/src/mod_privacy.erl
@@ -419,6 +419,27 @@ user_send_packet({#iq{type = Type,
 		false -> IQ
 	    end,
     {NewIQ, State};
+
+user_send_packet({#iq{type = Type,
+          from = #jid{luser = U, lserver = S},
+          to = #jid{luser = U, lserver = S},
+          sub_els = [_]} = IQ,
+        State})
+  when Type == set ->
+    case xmpp:get_subtag(IQ, #privacy_query{}) of
+      #privacy_query{active = undefined} -> {IQ, State};
+      #privacy_query{default = undefined, active = Active} ->
+        case get_user_list(U, S, Active) of
+          {ok, _} ->
+            % Adjust the client's state directly, so the next to-be-processed
+            % packet will take the active list into account.
+            {IQ, State#{privacy_active_list => Active}};
+          true ->
+            {IQ, State}
+        end;
+      _ -> {IQ, State}
+    end;
+
 user_send_packet(Acc) ->
     Acc.
 
-- 
2.19.1
```

nbruning:
That's strange, what is the crash you're seeing ?

nbruning:
Above error is fixed in last commit.

badlop:
Good, now the clients can login correctly. Have you tested the patch? With what client? In my tests with Tkabber, the function clause you added is not executed at all, because the IQ queries are handled by the first function clause. So, the presence stanza still passes to the contact.

badlop:
> The first clause only matches when a privacy list is already activated.

Then let's imagine:
1. a user has active any list, for example one called "block-unknown-users".
2. Later he sends your three stanzas, that define a new list "invisible",
3. and then make that list active
4. and finally send a presence.

That presence will pass to the destination because the first function clause processed step 3. The bug you were trying to fix in this commit is still present.

nbruning:
You're right, my patch was only working for the initial privacy list set.

I've refactored the patch. Updating of c2s state is now performed completely in `user_send_packet`, so the `user_receive_packet` hook is no longer needed.

badlop:
Running common tests with **./rebar skip_deps=true ct suites=ejabberd** fails in line https://github.com/processone/ejabberd/blob/master/test/privacy_tests.erl#L145

with
```
=== Ended at 2019-05-08 11:50:40
=== Location: [{privacy_tests,del_active,145},
              {test_server,ts_tc,1545},
              {test_server,run_test_case_eval1,1063},
              {test_server,run_test_case_eval,995}]
=== Reason: no match of right hand side value 
                 {privacy_query,[{privacy_list,<<"del-active">>,[]}],
                                none,<<"del-active">>}
  in function  privacy_tests:del_active/1 (privacy_tests.erl, line 145)
  in call from test_server:ts_tc/3 (test_server.erl, line 1545)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1063)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 995)
=== 
*** privacy_del_active failed.
    Skipping all other cases in sequence.
```

nbruning:
Hi @badlop,

Added extra clause for 'none, to unset privacy list, also rebased on current master branch.
CI seems to agree.

PR #2678, opened 2018-11-10
[FIX] iq_reply handle_cast to handle_info

songjungsoo:

   description.

   them as separate PRs.


zinid:
I wonder, if you're too lazy to write the PR description, why do you expect that we will read your patch?

PR #2696, opened 2018-11-22
Fix @from attribute on MucSub 'subscribers' node messages

Toons33:
According to [MucSub documentation](https://docs.ejabberd.im/developer/xmpp-clients-bots/extensions/muc-sub/#receiving-events), event messages are sent by the MUC room itself. This currently works for `messages` and `affiliation` nodes but not for the `subscribers` node.

When receiving events on the MUC subscribers node, the `from` attribute seems incorrect: instead of the room JID, it's the newly-subscribed JID.

```XML
<message from="charlie@server.com"
         to="anna@server.com/iphone">
   <event xmlns="http://jabber.org/protocol/pubsub#event">
     <items node="urn:xmpp:mucsub:nodes:subscribers">
       <item id="10776102417321261057">
         <subscribe xmlns="urn:xmpp:mucsub:0" jid="charlie@server.com" nick="Charlie"/>
       </item>
     </items>
   </event>
</message>
```

The values of `/message/@from` and `/message/event/items/item/subscribe/jid` are identical, and the MUC room's JID is not available in the stanza: **Anna** has no way of knowing which room **Charlie** just subscribed to.

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -4331,7 +4331,7 @@ send_subscriptions_change_notifications(From, Nick, Type, State) ->
 					items = [#ps_item{
 					    id = p1_rand:get_string(),
 					    sub_els = [Payload]}]}}]},
-			    ejabberd_router:route(xmpp:set_from_to(Packet, From, JID));
+			    ejabberd_router:route(xmpp:set_from_to(Packet, State#state.jid, JID));
 			false ->
 			    ok
 		    end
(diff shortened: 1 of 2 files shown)


Neustradamus:
@Toons33: Can you sign the CLA?

badlop:
Ah, this was fixed a year ago!
https://github.com/processone/ejabberd/commit/ea32d3f02c7a01a94998977c9bc1bf2d6d175777

What is brand new is your test, so I'll commit it, with a minor correction: the room sends the messages to the user bare JID, not the full JID.

PR #2703, opened 2018-12-02
Add 'xml' format to mod_muc_log

steffen-kiess:
Add 'xml' format to mod_muc_log

This format will write the MUC messages into an XML file. The file will
contain stanzas (`<message>` for the messages, <presence> for joining, leaving
and nick changes and `<iq type='set'>` for root configuration changes).
Every stanzas will contain a `<address>` element with the JID of the user
sending the message and a `<delay xmlns='urn:xmpp:delay'>` element with the
time of the stanza. All messages for a day will be in a
`<muclog xmlns='https://ejabberd.im/muclog'>` element.

### DIFF
--- a/src/mod_mam.erl
+++ b/src/mod_mam.erl
@@ -306,7 +306,7 @@ remove_mam_for_user_with_peer(User, Server, Peer) ->
     end.
 
 -spec get_room_config([muc_roomconfig:property()], mod_muc_room:state(),
-		      jid(), binary()) -> [muc_roomconfig:property()].
+		      undefined | jid(), binary()) -> [muc_roomconfig:property()].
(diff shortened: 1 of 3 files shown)


zinid:
And the point is? How this can be useful for ejabberd users?

steffen-kiess:
The advantage over the HTML format is that it is possible to change the way the MUC log is displayed later on (when converting the XML to HTML). With the HTML format the format is fixed when the log is written.

mremond:
I think the best way to handle this would be to make the MUC logger is a configurable API with hooks, and have an HTML logger as a standard module.
We would then be able to have contributed module doing the logging in different formats.

PR #2763, opened 2019-01-13
option to prevent archive removal on room destroy

ChaosKid42:
I'd like to keep the message archive even after a room is destroyed. Therefore I propose a new option of mod_mam `clear_archive_on_room_destroy` which defaults to `true`.

As this can lead to privacy issues (see weiss' comment below) I implemented a check, that disallows room creation if the archive for this newly created room is not empty. To be backwards compatible this check will only be performed if `clear_archive_on_room_destroy` is `false`.

My use-case is as follows: I'm mainly using private, persistent rooms. The archive is cleaned by regularily calling `ejabberdctl delete_old_messages <days>`. Rooms may be created and destroyed by users but I'd like to keep the room's message archives (until they are deleted by the above command) for reasons of accountability even if a user decides to destroy a room.

This PR also adds two new commands `ejabberdctl rooms_empty_list <Server>`  and `ejabberdctl rooms_empty_destroy <Server>` which allow to list and destroy rooms that have an empty archive.

I will happily write an entry for the documentation if my PR is accepted.

### DIFF
--- a/src/mod_mam.erl
+++ b/src/mod_mam.erl
@@ -41,7 +41,7 @@
 	 delete_old_messages/2, get_commands_spec/0, msg_to_el/4,
 	 get_room_config/4, set_room_option/3, offline_message/1, export/1,
 	 mod_options/1, remove_mam_for_user_with_peer/3, remove_mam_for_user/2,
-	 is_empty_for_user/2, is_empty_for_room/3]).
+	 is_empty_for_user/2, is_empty_for_room/3, check_create_room/4]).
 
 -include("xmpp.hrl").
 -include("logger.hrl").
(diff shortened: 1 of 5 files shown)


weiss:
This is hairy.  If, after room destruction, another user creates a room of the same name, he'll get access to the archive which may well have been restricted to the old room's members before.  I'd certainly not do this by default.

ChaosKid42:
@weiss: Right. That's why it's `true` by default. I will point out the risk of setting it to `false` in the documentation.

ChaosKid42:
@weiss: Maybe a check, that would reject room creation for rooms that still have archived contents would do. What do you think?
That would be similar to a tombstone.
Unfortunately I don't know enough erlang to do it...

weiss:
> I'd like to keep the room's message archives (until they are deleted by the above command) for reasons of accountability

Wouldn't it make sense to also add an such an option for user archives, by the way?  Except that this potentially introduces the same issue if a deleted account name is reused by someone else.  But maybe we could argue that this kind of account management issues are within the admin's realm.

ChaosKid42:
@weiss: Thanks for your advice. Maybe I'll find some time tonight to implement the changes.

ChaosKid42:
@weiss I believe I implemented the changes you suggested. What do you think?

PR #2764, opened 2019-01-13
acl for mam preferences

ChaosKid42:
I'd like to keep users from changing default mam policy. Therefore I propose a new option of mod_mam `access_prefernces` which defaults to true.

I will happily write an entry for the documentation if my PR is accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-# Version NEXT
+# Version 19.02
 
-*
+* New acl "access_preferences" for mam preferences
 
 # Version 18.12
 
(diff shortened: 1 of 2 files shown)


weiss:
To be honest, I'd turn this into an access rule (and call the option `access_preferences` or so).

Sorry, I'm not trying to annoy you :wink:

zinid:
I agree with @weiss: we definitely need to reuse existing ACL framework for things like this.

ChaosKid42:
I see your point. I'll try to work in that direction maybe by end of this week.

ChaosKid42:
@weiss Maybe now it is acceptable.

weiss:
[Merged][1], thanks.

[1]: https://github.com/processone/ejabberd/commit/fb068557b29a870390ea66170566766e8a96fca9

PR #2781, opened 2019-01-29
Fix ping IQ reply/timeout processing ("mod_ping" regression since 17.…

fdie:
Since the 17.x, a new IQ handling module "ejabberd_iq" has been created, the results/timeout of a submitted IQ is no longer sent to the requesting process with "gen_server:cast" but with an erlang send method (!). Thus the "iq_reply" response should be handled in "handle_info" rather than "handle_cast" methods.

The consequence is that sessions are never closed even if they don't reply to pings...

### DIFF
--- a/src/ejabberd_http_ws.erl
+++ b/src/ejabberd_http_ws.erl
@@ -108,8 +108,9 @@ close({http_ws, FsmRef, _IP}) ->
 reset_stream({http_ws, _FsmRef, _IP} = Socket) ->
     Socket.
 
-change_shaper({http_ws, FsmRef, _IP}, Shaper) ->
-    p1_fsm:send_all_state_event(FsmRef, {new_shaper, Shaper}).
+change_shaper({http_ws, _FsmRef, _IP}, _Shaper) ->
+    %% TODO???
+    ok.
 
 get_transport(_Socket) ->
     websocket.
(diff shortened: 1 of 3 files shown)


fdie:
I missed to create 2 branches for the "mod_ping" / "websockets shaping" patches :-(
That's why I reverted some commits on this pull request too, but it should be fine for merging.

To be complete, I guess the sending in "ejabberd_iq:callback" have been converted to a "!" rather than a "gen_server:cast" since the requesting Proc could be anything else than a "gen_server" (an fsm for instance).

However the Proc may be invalid that raises a "badarg" exception, thus pull request should be completed with :

diff --git a/src/ejabberd_iq.erl b/src/ejabberd_iq.erl
index 8c731e0a..aeaffccd 100644
--- a/src/ejabberd_iq.erl
+++ b/src/ejabberd_iq.erl
@@ -173,4 +173,8 @@ calc_checksum(Data) ->
 callback(undefined, IQRes, Fun) ->
     Fun(IQRes);
 callback(Proc, IQRes, Ctx) ->
-    Proc ! {iq_reply, IQRes, Ctx}.
+    try
+        Proc ! {iq_reply, IQRes, Ctx}
+    catch _:badarg ->
+        ok
+    end.

I put everything in a dedicated branch "fdie:mod_ping_no_kill"

PR #2821, opened 2019-03-11
Origin header validation on websocket connection

pdorofiejczyk:
At present, ejabberd is not protected against CSRF attacks on websocket connection. This PR adds optional validation for ```Origin``` header to protect against connections from other domains than given in config (```websocket_origin``` option). This kind of protection can be provided by layer7 load balancer (ex. haproxy + lua), but, thanks to this PR, the lower layer load balancer can get chosen for a specific ejabberd implementation while still providing a secure websocket connection.

### DIFF
--- a/src/ejabberd_http_ws.erl
+++ b/src/ejabberd_http_ws.erl
@@ -371,5 +371,7 @@ opt_type(websocket_ping_interval) ->
     fun (I) when is_integer(I), I >= 0 -> I end;
 opt_type(websocket_timeout) ->
     fun (I) when is_integer(I), I > 0 -> I end;
+opt_type(websocket_origin) ->
+    fun (O) -> O end;
 opt_type(_) ->
-    [websocket_ping_interval, websocket_timeout].
+    [websocket_ping_interval, websocket_timeout, websocket_origin].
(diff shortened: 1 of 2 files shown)


bhaku:
👍

prefiks:
As optional feature it looks ok to me, so let's merge this!

Thank you for your contribution.

badlop:
I guess that should be documented in the ejabberd Guide. Is that paragraph correct?

```
- **`websocket_origin: ignore|URL`**: This option enables validation for `"Origin"` header
to protect against connections from other domains than given in config.
In this way, the lower layer load balancer can get chosen for a specific
ejabberd implementation while still providing a secure websocket connection.
Default value of this option is `ignore`.
Example value of URL: `"https://test.example.org:8081"`
```

On the other hand, shouldn't the value be checked and converted to binary in opt_type?

pdorofiejczyk:
@badlop The paragraph you provided looks correct.

Regarding to the value convertion it looks like possible improvement, but not critical.

zinid:
> As optional feature it looks ok to me, so let's merge this!

This is not an optional feature - ejabberd requires it to be presented. However, this violates RFC 6455:
> The request MUST include a header field with the name |Origin| [RFC6454] if the request is coming from a browser client.  If the connection is from a non-browser client, the request MAY include this header field if the semantics of that client match the use-case described here for browser clients.

We must change the code to not require the header if it's not configured. Since I have no time to mess with that I will revert the merge if the fix is not provided ASAP.

pdorofiejczyk:
Ok. I'll try to fix this ASAP.

pdorofiejczyk:
@zinid Please check PR #2868 if this fixes problem you have mentioned above.

PR #2838, opened 2019-03-21
SQL for pgsql - set family in marks

genofire:
It is reserved by the cockroachdb implementation: https://www.cockroachlabs.com/docs/v2.1/column-families.html#main-content

Not importent, I am just playing a little bit ...

### DIFF
--- a/sql/pg.new.sql
+++ b/sql/pg.new.sql
@@ -282,7 +282,7 @@ CREATE TABLE vcard_search (
     server_host text NOT NULL,
     fn text NOT NULL,
     lfn text NOT NULL,
-    family text NOT NULL,
+    "family" text NOT NULL,
     lfamily text NOT NULL,
     given text NOT NULL,
     lgiven text NOT NULL,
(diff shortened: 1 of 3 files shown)


genofire:
Okay there would be some sql parse fixes on the vcard module necessary.

zinid:
> Okay there would be some sql parse fixes on the vcard module necessary.

Why, cockroachdb doesn't accept the request without those quotes?

genofire:
because `family` is a reserved name in this postgresql implementation (see link in description of this PR).
(if i am correct, it is kind of index-like database speedup)

zinid:
@licaon-kter I am not sure this is fully fixed. I see in the PR that there are quotes around `family` in the SQL request itself. Is it really needed? I personally don't know.

genofire:
yes it is needed, but i do not find all queries .....

zinid:
@genofire yeah, this is bad...

zinid:
Closed in f51ba687e213cdc4136138cba39bb8bc022097ab
However, there are still a few unresolved issues left regarding CockroachDB compatibility.

PR #2868, opened 2019-04-24
Fix #2821 RFC6454 violation on websocket connection when validating Origin

pdorofiejczyk:
This PR fixes RFC6454 introduced in #2821


zinid:
The patch is very elegant, thanks a lot for the fast reaction.

PR #2869, opened 2019-04-25
Fix #2540 Ejabberd doesnt compile as mix umbrella project dependency

### LINKED ISSUE #2540: Ejabberd doesn't compile as mix umbrella project dependency

18.04

Elixir docker image

hex package

I have created basic mix umbrella project [https://github.com/pdorofiejczyk/elixir_umbrella_hello_world](https://github.com/pdorofiejczyk/elixir_umbrella_hello_world). It contains one elixir application which have ejabberd as dependendency. When I'm trying to compile, I'm getting error ```src/mod_muc_log.erl:47: can't find include file "xmpp.hrl"```. Everything works fine if it is not umbrella project.

pdorofiejczyk:
This PR fixes #2540. It is not perfect way but it works for most of cases. The problem remains when given project has custom ```deps_path``` configuration but as I know there is no way to get this config option from dependencie's ```mix.exs``` level.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -96,9 +96,13 @@ defmodule Ejabberd.Mixfile do
   end
 
   defp deps_include(deps) do
-    base = case Mix.Project.deps_paths()[:ejabberd] do
-      nil -> "deps"
-      _ -> ".."
+    base = if Mix.Project.umbrella?() do
+      "../../deps"
+    else
+      case Mix.Project.deps_paths()[:ejabberd] do
+        nil -> "deps"
+        _ -> ".."
+      end
     end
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -96,9 +96,13 @@ defmodule Ejabberd.Mixfile do
   end
 
   defp deps_include(deps) do
-    base = case Mix.Project.deps_paths()[:ejabberd] do
-      nil -> "deps"
-      _ -> ".."
+    base = if Mix.Project.umbrella?() do
```
fabiankrol: 👏

mremond:
Many thanks for the patch!

PR #2896, opened 2019-05-30
Fix typo

licaon-kter:


zinid:
The upcoming "new configurator" branch will get rid of this warning (actually it will be removed to another file, slightly modified). So I close the PR. Thanks nonetheless.

Issue #2933, opened 2019-07-08
mod_muc: add additional hooks


Slawutich:
**Is your feature request related to a problem? Please describe.**
I want to implement 2 modules
1. block presence and message forks, by condition (for example, displayed messages, if there are more participants in the room than X)
2. automatically add hidden bots (XEP-0114) to the room

**Describe the solution you'd like**
The `mod_muc_room` hooks allow filtering packets at the entrance.
to implement what I need, I would like to filter the packets at the output before `ejabberd_router:route`

Slawutich:
Do you need more information from me?

mremond:
No, thanks. I do not need more information.

zinid:
Travis actually fails.

zinid:
Ah, seems like internal Travis error:
```
failed with return code 128 and the following output:
Cloning into 'cache_tab'...
fatal: unable to access 'https://github.com/processone/cache_tab/': Failed to connect to github.com port 443: Connection timed out

ERROR: 'get-deps' failed while processing /home/travis/build/processone/ejabberd: rebar_abort
Makefile:102: recipe for target 'deps/.got' failed
make: *** [deps/.got] Error 1
```

zinid:
By the way, the commit violates [Coding style convention](https://docs.ejabberd.im/developer/guide/#coding-style-convention).

mremond:
We have already features that limit the broadcast of presence and hiding so users presence is not a compelling use case. We do not want to add more hooks if they are not generally useful, so I do not think we should had those to ejabberd.

PR #2934, opened 2019-07-08
mod_muc: add additional hooks #2933

Slawutich:
mod_muc: add additional hooks #2933


mremond:
We have already features that limit the broadcast of presence and hiding so users presence is not a compelling use case. We do not want to add more hooks if they are not generally useful, so I do not think we should had those to ejabberd.

As a side comment, please note that both test suite and coding conventions were broken by this patch.

Issue #2965, opened 2019-07-26
mod_multicast [18.09] - Local limits disco error "Hook c2s_handle_info crashed"


McPo:
When specifying local limits in the ejabberd config for mod_multicast, if another component (jicofo) attempts to disco the multicast service, it crashes. (Whether jicofo should disco multicast is up for debate, however the service shouldn't crash just because local limits are configured)

It in fact appears that default values are not added to the disco result, but user defined values are, and as jicofo is local, only defining the local parameters causes the error.

Version: 18.09

Healthy Config
```
mod_multicast:
    limits:
        local: {}
        remote:
            message: 20
            presence: 20
```

Unhealthy Config
```
mod_multicast:
    limits:
        local:
            message: 100
            presence: 100
        remote:
            message: 20
            presence: 20
```

Healthy Log
```
[36mxmpp_1           |[0m 11:54:54.131 [info] (tcp|<0.691.0>) Received XML on stream = <<"<a xmlns='urn:xmpp:sm:3' h='50'/><iq to='multicast.docker-ec2.xxxxxxx.com' id='tgRjd-2816' type='get'><query xmlns='http://jabber.org/protocol/disco#info'></query></iq>">>
[36mxmpp_1           |[0m 11:54:54.131 [info] on_message [from:{jid,<<"focus">>,<<"docker-ec2.xxxxxxx.com">>,<<"focus5433138606317438">>,<<"focus">>,<<"docker-ec2.xxxxxxx.com">>,<<"focus5433138606317438">>}, to:{jid,<<>>,<<"multicast.docker-ec2.xxxxxxx.com">>,<<>>,<<>>,<<"multicast.docker-ec2.xxxxxxx.com">>,<<>>}, p:{iq,<<"tgRjd-2816">>,get,<<"en">>,{jid,<<"focus">>,<<"docker-ec2.xxxxxxx.com">>,<<"focus5433138606317438">>,<<"focus">>,<<"docker-ec2.xxxxxxx.com">>,<<"focus5433138606317438">>},{jid,<<>>,<<"multicast.docker-ec2.xxxxxxx.com">>,<<>>,<<>>,<<"multicast.docker-ec2.xxxxxxx.com">>,<<>>},[{xmlel,<<"query">>,[{<<"xmlns">>,<<"http://jabber.org/protocol/disco#info">>}],[]}],#{ip => {0,0,0,0,0,65535,49320,61450}}}]
[...]

PR #2968, opened 2019-07-29
mod_multicast service discovery bugfix

McPo:
Resolves the issue detailed in https://github.com/processone/ejabberd/issues/2965. 
Essentially c2s_handle_info crashes when specifying limits in the configuration, during a service discovery.

I haven't tested this on master (18.09) but the problematic code appears to still be present master.


badlop:
Thanks for reporting and fixing!

I've committed a slitghly modified version, so the function returns always a list:
https://github.com/processone/ejabberd/commit/630cfe26b7c460d9860a27eee215cd70aaf47a0d

PR #3002, opened 2019-08-22
discover info fix mod_multicast

jaspreeterlang:
<iq to='multicast.host' id='kVW7v-22' type='get'><query xmlns='http://jabber.org/protocol/disco#info'></query></iq>

This discovery was not returning anything.

   description.
No, I have not seen issue reported, I faced it myself.

Yes, I have.

   them as separate PRs.
It's Atomic.

4. If you change or new feature involves storage backends, did you make
No changes in backend.

I have given example of how to test.

Not needed.


jaspreeterlang:
> Hello,
> 
> Are you sure that change is needed, it seems that iq_disco_info_extras is already returning `list(#xdata{})` so i don't think second wrapping in list in needed here?

Actually, I was on 18.09 version and it gave me an issue. I just noticed new master have a fix for it in a diffent way(iq_disco_info_extras). I will close this PR.

PR #3048, opened 2019-10-02
Fixing an error when call remove mam from xmlrpc

alamse:
This PR is to fixing an error when we tried to call `remove_mam_for_user` and `remove_mam_for_user_with_peer` from xmlrpc.

```
CRASH REPORT Process <0.1308.0> with 0 neighbours crashed with reason: no function clause matching lists:flatten(<<"MAM archive removed">>) line 616
```

### DIFF
--- a/src/ejabberd_xmlrpc.erl
+++ b/src/ejabberd_xmlrpc.erl
@@ -371,7 +371,7 @@ format_result(Code, {Name, rescode}) ->
 format_result({Code, Text}, {Name, restuple}) ->
     {struct,
      [{Name, make_status(Code)},
-      {text, lists:flatten(Text)}]};
+      {text, io_lib:format("~ts", [Text])}]};
 format_result(Elements, {Name, {list, ElementsDef}}) ->
     FormattedList = lists:map(fun (Element) ->
 				      format_result(Element, ElementsDef)
--- a/src/mod_mam.erl
+++ b/src/mod_mam.erl
@@ -312,11 +312,11 @@ remove_mam_for_user(User, Server) ->
     Mod = gen_mod:db_mod(LServer, ?MODULE),
     case Mod:remove_from_archive(LUser, LServer, none) of
 	ok ->
-	    {ok, "MAM archive removed"};
+	    {ok, <<"MAM archive removed">>};
 	{error, Bin} when is_binary(Bin) ->
 	    {error, Bin};
 	{error, _} ->
-	    {error, "Db returned error"}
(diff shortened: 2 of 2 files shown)


badlop:
There are other commands that also return binary in restuple. That isn't a problem for other command frontends like ejabberdctl and mod_http_api, so I think the problem is not in mod_mam commands; the problem is in ejabberd_xmlrpc that doesn't know how to handle binary restuple.

Try this patch instead:
```diff
diff --git a/src/ejabberd_xmlrpc.erl b/src/ejabberd_xmlrpc.erl
index afbe61f34..6e67dc4c4 100644
--- a/src/ejabberd_xmlrpc.erl
+++ b/src/ejabberd_xmlrpc.erl
@@ -371,7 +371,7 @@ format_result(Code, {Name, rescode}) ->
 format_result({Code, Text}, {Name, restuple}) ->
     {struct,
      [{Name, make_status(Code)},
-      {text, lists:flatten(Text)}]};
+      {text, io_lib:format("~ts", [Text])}]};
 format_result(Elements, {Name, {list, ElementsDef}}) ->
     FormattedList = lists:map(fun (Element) ->
 				      format_result(Element, ElementsDef)
```

alamse:
> There are other commands that also return binary in restuple. That isn't a problem for other command frontends like ejabberdctl and mod_http_api, so I think the problem is not in mod_mam commands; the problem is in ejabberd_xmlrpc that doesn't know how to handle binary restuple.
> 
> Try this patch instead:
> 
> ```diff
> diff --git a/src/ejabberd_xmlrpc.erl b/src/ejabberd_xmlrpc.erl
> index afbe61f34..6e67dc4c4 100644
> --- a/src/ejabberd_xmlrpc.erl
> +++ b/src/ejabberd_xmlrpc.erl
> @@ -371,7 +371,7 @@ format_result(Code, {Name, rescode}) ->
>  format_result({Code, Text}, {Name, restuple}) ->
>      {struct,
>       [{Name, make_status(Code)},
> -      {text, lists:flatten(Text)}]};
> +      {text, io_lib:format("~ts", [Text])}]};
>  format_result(Elements, {Name, {list, ElementsDef}}) ->
>      FormattedList = lists:map(fun (Element) ->
>  				      format_result(Element, ElementsDef)
> ```

thanks you @badlop , i will try it.

alamse:
@badlop i've tried it, you are right. new commit has include that changes.

badlop:
Ok, I've applied my small patch which solves the problem.

PR #3083, opened 2019-11-14
Use a numeric value for loglevel

marc0s:
`loglevel` accepts numeric values (see https://docs.ejabberd.im/admin/configuration/#logging), not log level's names.


zinid:
The example config in `master` branch is supposed to be used with `master` branch, where loglevels have been changed to literal values. If you're using older version, then switch to the corresponding tag and use the example config from this tagged version.

PR #3096, opened 2019-11-26
Update dependencies usages to make it possible to use ejabberd as rebar3 dependency.

kappac:
Use 'include_lib' instead of 'include' for 'fast_xml', 'xmpp', 'p1_utils' to make possible use 'ejabberd' as a rebar3 dependency.


prefiks:
Hello,
If i remember correctly we did switch to using extra include dirs instead of include_lib because there was some problem with loading external modules. I will need to check if this commit doesn't broke this part or if does to see if there is better workaround for this issue.
 In general i think your changes are correct thing to do, but we will see if everything work as expected.

weiss:
Thanks to @nosnilmot, building with Rebar3 should be fixed now.

PR #3103, opened 2019-12-02
Fix "ejabberd_iq" processing of the IQ id for proc delivery

fdie:
"ejabberd_iq" seems to be an IQ handler for internal usage (see "mod_ping") where the routing is based on the information encoded in the IQ id field ('rr-').

The last field of such an "rr-" id is the hash of the node.
"ejabberd_iq" uses "ejabber_cluster:node_id" for the encoding & "ejabberd_cluster:get_node_by_id" for the decoding of this field.

Even if the "node_id()" method has no Pid parameter, I guess the Proc provided in the "ejabberd_iq:route" method must be a process on the executing node.

If my assumption is right, there is a close link between the Node and the Proc and if the Node information could not be decoded, there is no chance to seek any process to deliver the IQ result...

In fact, the "ejabberd_iq:decode_id" does it right by catching a {badmatch, xxx} exception and proceed with standard routing when no Node could be matched, perfect.

decode_id(<<"rr-", ID/binary>>) ->
    try
      ...
    catch _:{badmatch, _} ->
	    error
    end;

Miserably, "ejabberd_cluster" never throws such an exception, but returns any node on the current cluster !?

With the following fix, a Proc is able to MAM query a room hosted on a remote cluster through XMPP federation (S2S).

### DIFF
--- a/src/ejabberd_cluster_mnesia.erl
+++ b/src/ejabberd_cluster_mnesia.erl
@@ -108,7 +108,7 @@ get_node_by_id(Hash) ->
     try binary_to_integer(Hash) of
 	I -> match_node_id(I)
     catch _:_ ->
-	    node()
+        throw({badmatch, Hash})
     end.
 
 -spec send({atom(), node()}, term()) -> boolean().
@@ -150,5 +150,5 @@ match_node_id(I, [Node|Nodes]) ->
 	I -> Node;
 	_ -> match_node_id(I, Nodes)
     end;
-match_node_id(_I, []) ->
-    node().
+match_node_id(I, []) ->
+    throw({badmatch, I}).
--- a/src/mod_proxy65_service.erl
+++ b/src/mod_proxy65_service.erl
@@ -214,7 +214,11 @@ process_bytestreams(#iq{type = set, lang = Lang, from = InitiatorJID, to = To,
     ACL = mod_proxy65_opt:access(ServerHost),
     case acl:match_rule(ServerHost, ACL, InitiatorJID) of
 	allow ->
-	    Node = ejabberd_cluster:get_node_by_id(To#jid.lresource),
+	    Node = try
+		       ejabberd_cluster:get_node_by_id(To#jid.lresource)
+		   catch _:{badmatch, _} ->
+		       node()
+		   end,
 	    Target = jid:encode(jid:tolower(TargetJID)),
 	    Initiator = jid:encode(jid:tolower(InitiatorJID)),
 	    SHA1 = str:sha(<<SID/binary, Initiator/binary, Target/binary>>),


zinid:
What's the point in throwing an exception? So admins will come to our bugtracker creating pointless issues?

fdie:
As said in "ejabberd_iq:decode_id" there is a catch, so this exception was expected...
But wait, I just saw that tests are now are failing, let me check why...

fdie:
Sorry, I missed this new use in "src/mod_proxy65_service.erl", will have to change my pull-request

fdie:
I just added a catch in "mod_proxy65_service:process_bytestreams" as you did in "ejabberd_iq".
But you are right, throwing an exception could be replaced by "{error, nomatch}"

-spec get_node_by_id(binary()) -> node() | {error, nomatch}

Either way, this method could fail and the current implementation that returns the current node when the id doesn't match does not make sense, does it ?

zinid:
I don't think we need to change the API of ejabberd_cluster backends. Because that will require changing for example the code of ejabberd Business Edition (which uses a different cluster backend, not `ejabberd_cluster_mnesia`). Also this API function is used in other parts of eBE. If you really want to log something on node id mismatch in ejabberd_iq you should do it there. Although I suggest doing this in debug level, because a large node failure may lead to tons of pointless log entries.

fdie:
I understand your reticence but this API is definitely weird ;-)

% returns the NODE with the provided HASH and if no node matches, just return the "current node"
-spec get_node_by_id(binary()) -> node()

fdie:
If you reject the pull request, you may at least remove the catch of “nomatch“ in ejabberd_iq because it is never thrown :-)

zinid:
Do you mean 'badmatch'? Of course it will be thrown if a client provides invalid value of 'rr'.

fdie:
Sorry, you are right the checksum causes a badmatch for a wrong 'rr-' value and I wasn't that clear, let me be more precise.
My concern is more about the 'ejabberd_cluster_mnesia:get_node_by_id(...)' returned value usefullness.
I thought the 'badmatch' should also be raised when the hash doesn't match any node of the cluster... and I still think for the 'ejabberd_iq' usage it makes sense to return an error in any form but you know the overall usage for sure.

The ultimate reason (see 'ejabberd_iq:dispatch(...)' to route the IQ result back to one and only the one identified node is because it's the only node containing the 'ets' stored information for the proper IQ result processing.
That's why, to my understanding, if the node doesn't match any node (see 'ejabberd_cluster_mnesia:match_node_id(...)'), it is useless to return any node (even the current node as currently) because 'ejabberd_iq:dispatch(...)' may very often do not execute on the node containing the right 'ets' table in an eJabberd cluster environment.

I agree, this bug is minor, and it should not happen. Maybe this use case may help: "on a node N, an 'ejabberd_iq:route' is executed, some information for the result dispatch is stored in an 'ets' table, then the node disappears from the cluster, consequently no more node matches the hash of the 'rr-' value and the result will be processed by some default node which is useless since the 'ets' table has gone and any proc delivery will fail since the the node N is no longer available..."

As said the fix solves my use case for S2S remote IQ processing, which is currently impossible with 'ejabberd_iq' since the result processing is attempted on the remote cluster and is not sent back to the originating cluster for proper processing there. But maybe this is not a feature ;-)

HTH, thanks for your time

fdie:
Hi, I finally made this change to support "XMPP S2S federation" with "ejabberd_iq" sent IQs because remote nodes can't match the "node_id" encoded in the the "rr-" IQ id for sure ;-)

% return "undefined" if the node does not match, let the API client decide about any valid fallback node...

-spec get_node_by_id(binary()) -> node() | undefined.

see https://github.com/fdie/ejabberd/commit/defe3b84e47a1c1888ad1560a7664752461e0e2c

zinid:
> As said the fix solves my use case for S2S remote IQ processing, which is currently impossible with 'ejabberd_iq' since the result processing is attempted on the remote cluster and is not sent back to the originating cluster for proper processing there.

I didn't understand anything. What is "the remote cluster"?

zinid:
I also fail to see how your patch fixes anything at all: if the response is processed on incorrect node, ETS lookup would fail and nothing will happen: exactly the same situation as we just return 'error' and don't process anything at all.

fdie:
You are completely right, indeed on a cluster it doesn't change anything ;-)
Here is my use case (3rd party service) and it solves it definitely.

We have 2 eJabberd clusters 'domain.ru' & 'domain.eu'.
A module running in the 'domain.ru' cluster is able to send an IQ request to a room for instance running on the 'domain.eu' cluster using the 'ejabberd_iq' service.

- without the fix, the IQ with 'rr-xxx' generated on the 'domain.ru' cluster is routed through S2S to the 'domain.eu' cluster but the result is never returned because the 'ejabberd_iq' on "domain.eu" dispatches the stanza to any node even if the nodeId doesn't match :-(

- with the fix, since the nodeId doesn't match any node on the 'domain.eu' cluster, it is returned back through S2S to the 'domain.ru' cluster and then correctly routed to the requesting proc :-)

zinid:
> We have 2 eJabberd clusters 'domain.ru' & 'domain.eu'.

I stopped reading after this. Seems like you're messing with the definitions.

PR #3108, opened 2019-12-09
Add Hungarian translation, update existing translations

urbalazs:
Added Hungarian language file, and regenerate existing translations with command `make translations`.


badlop:
Thanks! I've updated first the existing files, and then added your new translation.

PR #3125, opened 2019-12-24
enable odbc in mix build

ChaosKid42:
I'd like to use the official ejabberd docker image with odbc database connectivity. Therefore this change is necessary in addtition to a change in docker-ejabberd:

See https://github.com/processone/docker-ejabberd/issues/50 and https://github.com/processone/docker-ejabberd/pull/51.


mremond:
Thanks. Sorry we did not merge it earlier.
That patch makes sense.

PR #3127, opened 2019-12-27
allow storage of emojis in archive on mssql

ChaosKid42:
I found that ejabberd doesn't properly store emojis in mssql. Reason is given here: https://www.freetds.org/userguide/localization.htm

I tested this PR with ejabberd 19.09.74 (self-compiled) and following version of db:

Microsoft SQL Server 2019 (RTM-GDR) (KB4517790) - 15.0.2070.41 (X64) 
	Oct 28 2019 19:56:59 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS) <X64>


mremond:
@prefiks This PR looks good to me.
What do you think ?

prefiks:
I have no way to test it, but looking at mssql docs, it should be working. So i say we can merge it.

Issue #3182, opened 2020-03-05
admin command rooms_empty_destroy does not destroy any rooms


ChaosKid42:
## Environment

- ejabberd version: HEAD
- Erlang version: Erlang (SMP,ASYNC_THREADS,HIPE) (BEAM) emulator version 10.6.4
- OS: Arch Linux
- Installed from: source

## Errors from error.log/crash.log

No errors

## Bug description

Due to a mistake in mod_muc_admin.erl rooms_empty_destroy does not identify empty rooms and therefore cannot delete them. Fix: https://github.com/processone/ejabberd/pull/3183

PR #3183, opened 2020-03-05
fix admin command rooms_empty_destroy

ChaosKid42:
This is supposed to fix https://github.com/processone/ejabberd/issues/3182


prefiks:
Looks fine, thank you for noticing this.

PR #3192, opened 2020-03-17
Fixed HTTP PUT body parsing error

gihook:
Think I found a bug in ejabberd_http module. Process (process) function called from HTTP handler module trims body for large PUT requests (it seems that this depends on length of header). I guess that PUT can be handled the same way as POST, when this patch is applied Ejabberd correctly parses the body.

### DIFF
--- a/src/ejabberd_http.erl
+++ b/src/ejabberd_http.erl
@@ -400,21 +400,16 @@ extract_path_query(#state{request_method = Method,
 	    ?DEBUG("Error decoding URL '~p': ~p", [Path, Error]),
 	    {State, false};
         {LPath, _Query} ->
-	    case Method of
-		'PUT' ->
-		    {State, {LPath, [], Trail}};
-		'POST' ->
-		    case recv_data(State) of
-			{ok, Data} ->
-			    LQuery = case catch parse_urlencoded(Data) of
-					 {'EXIT', _Reason} -> [];
-					 LQ -> LQ
-				     end,
-			    {State, {LPath, LQuery, Data}};
-			error ->
-			    {State, false}
-		    end
-	    end
+            case recv_data(State) of
+                {ok, Data} ->
+                    LQuery = case catch parse_urlencoded(Data) of
+                                 {'EXIT', _Reason} -> [];
+                                 LQ -> LQ
+                             end,
+                    {State, {LPath, LQuery, Data}};
+                error ->
+                    {State, false}
+            end
     end;
 extract_path_query(State) ->
     {State, false}.


badlop:
The different behaviours depending on the method was implemented back in 063737e4f5158a52abde4388fa51a6a68ddcae51, so I guess that isn't a mistake.

gihook:
> The different behaviours depending on the method was implemented back in [063737e](https://github.com/processone/ejabberd/commit/063737e4f5158a52abde4388fa51a6a68ddcae51), so I guess that isn't a mistake.

Any idea for different fix?

weiss:
> Any idea for different fix?

We should try to understand the actual problem before thinking about a fix.  As far as I know, the only code that accepts `PUT` requests within ejabberd is `mod_http_upload`, which handles large files just fine (for me).  Are you using `ejabberd_http` from a custom module?

gihook:
Yes, I'm using custom elixir module to handle this request. I haven't managed to reproduce exact error (for me the body is only partially trimmed), but I've reproduced similar error (in this example body is not parsed at all, i.e. it is empty)

changes to reproduce error:
https://github.com/gihook/ejabberd/pull/1/files

ejabberd is started via iex -S mix, curl-test.sh is used to create HTTP request

weiss:
Sounds like you're simply not calling [`:ejabberd_http.recv_file(request, path)`][1] to receive the body.

[1]: https://github.com/processone/ejabberd/blob/20.02/src/ejabberd_http.erl#L571

gihook:
:ejabberd_http.recv_file(request, path) writes body to disc, but I would like to use body without writing it. I would expect body to be parsed in request record (as it is for POST requests).

Sometimes body is correctly parsed, example request:
curl 'http://localhost:5280/test' -X PUT  -H 'accept: application/json'  -H 'content-type: application/json'  --data-binary '{"test": 1}'

Is there a configurable limit for max body size?

weiss:
> :ejabberd_http.recv_file(request, path) writes body to filesystem, but I would like to use body without writing it. I would expect body to be parsed in request record (as it is for POST requests).

`ejabberd_http` is tailored to ejabberd's own requirements, it's not meant to be a generic HTTP server.  For non-trivial purposes, you might want to use a separate Erlang/Elixir HTTP server.  As a quick hack for using `ejabberd_http` you could write the body to some memory filesystem (e.g. `/dev/shm` on typical Linux systems), of course.

> Sometimes body is correctly parsed

Yes, `ejabberd_http` reads the first chunk of data and that might or might not include the full body.  It no longer reads arbitrary amounts of data into memory as that was an easy way to DoS the server.

> Is there a configurable limit for max body size?

You can't tell `ejabberd_http` to read a larger chunk of data into memory, no.

PR #3202, opened 2020-03-29
sql_type should be taken for LServer, not LHost

borischernov:
sql_type option should be retrieved for the main domain, not the MUC subdomain


prefiks:
Hello,

It's seems to be correct, well spotted, i will merge this, thanks!

Issue #3206, opened 2020-04-02
Show nick in MucSub


snoopcatt:
**Is your feature request related to a problem? Please describe.**
I'm trying to implement MUC subscribers list with MucSub spec.
But when admin or moderator requests subscriber list (or public member list is enabled in room settings), there is no subscribers nicks in response, only their JID. 

**Describe the solution you'd like**
So, as nick is unique (it is impossible to join with occupied by some subscriber nickname), I suggest always return nick along with JID even if we have rights to see JID.

**Describe alternatives you've considered**
There is no alternate ways to get subscribers list with their nicknames.

**Additional context**
There is a simple patch to deal with it:

```
--- ejabberd-20.02.orig/src/mod_muc_room.erl
+++ ejabberd-20.02/src/mod_muc_room.erl
@@ -4306,7 +4306,7 @@ process_iq_mucsub(From, #iq{type = get,
                     fun(_, #subscriber{jid = J, nick = N, nodes = Nodes}, Acc) ->
                         case ShowJid of
                             true ->
-                                [#muc_subscription{jid = J, events = Nodes}|Acc];
+                                [#muc_subscription{jid = J, nick = N, events = Nodes}|Acc];
                             _ ->
                                 [#muc_subscription{nick = N, events = Nodes}|Acc]
                         end
```

Neustradamus:
@snoopcatt: Can you sign the CLA on your PR?
Badly it was not included in 20.04 :/

PR #3207, opened 2020-04-02
Always show MucSub subscribers nicks (#3206)

snoopcatt:
Related to https://github.com/processone/ejabberd/issues/3206
Changes behaviour of: https://docs.ejabberd.im/developer/xmpp-clients-bots/extensions/muc-sub/#getting-list-of-subscribers-of-a-room

Always return subscriber nick in reply to get subscribers list, even if we have rights to see full JID.


Neustradamus:
@snoopcatt: Can you sign the CLA?

PR #3218, opened 2020-04-09
Bump jiffy so it compiles on older GCC

licaon-kter:
...like 4.4.7 used in RH/CentOS 6

FYI I didn't bump it in https://github.com/processone/ejabberd/blob/master/mix.lock since I don't know where that hash comes from exactly.


licaon-kter:
Any blockers?

prefiks:
Thank you for that, i will merge it.

PR #3235, opened 2020-04-24
Support STUN/TURN service discovery

### LINKED ISSUE #2947: External Service Discovery
**Is your feature request related to a problem? Please describe.**
Several (if not all) WebRTC clients, like NextCloud Talk or JavaScript XMPP Chat, ask for a TURN secret to authenticate but Ejabberd does not seem to setup one (correct me if I'm wrong).

**Describe the solution you'd like**
I would like a parameter for listening module `ejabberd_stun` to set a TURN REST API secret.
I'm guessing `auth_type` should also be updated to allow a `secret` type.
Configuration could look like that:
```
  -
    port: 3478
    ip: "0.0.0.0"
    transport: udp
    module: ejabberd_stun
    use_turn: true
    turn_ip: "10.20.30.40"
    auth_type: secret
    auth_secret: "mysupersecretsharedwithstunturnclient"
    auth_realm: "example.com"
```

As reference, same function but from the [coturn doc](https://github.com/coturn/coturn/wiki/turnserver):
> --use-auth-secret TURN REST API flag. Flag that sets a special WebRTC authorization option that is based upon authentication secret. This feature purpose is to support "TURN Server REST API" as described in TURNServerRESTAPI.pdf in the docs. This option is used with long-term credentials mechanism.
> --static-auth-secret <secret> Static authentication secret value (a string), for TURN REST API. If not set, then the turn server will try to use the 'dynamic' value in turn_secret table in user database (if present). The database-stored value can be changed on-the-fly by a separate program, so this is why that other mode is 'dynamic'. Multiple shared secrets can be used (both in the database and in the "static" fashion).
> https://github.com/coturn/coturn/wiki/turnserver#webrtc-usage

**Describe alternatives you've considered**
Not using ejabberd for STUN/TURN ? 😅 

**Additional context**
NA

weiss:
Add the `mod_stun_disco` module, which allows XMPP clients to discover STUN/TURN services and to obtain temporary credentials for using them as per [XEP-0215: External Service Discovery][1].  The upcoming A/V support in the Android client Conversations will hard-depend on this server feature, and it's also used by clients such as Movim (and Jitsi Meet, though they [currently use an outdated version][2] of XEP-0215).

The format of the temporary credentials handed out to clients is described in an [IETF draft][3].  Therefore, the module cannot only be used with ejabberd's built-in STUN/TURN support, but also with external STUN/TURN servers that support the same draft ([such as coturn][4] or restund).  It also allows (non-XMPP) WebRTC applications to use ejabberd's built-in STUN/TURN support.

This PR doesn't touch existing code except for:

- making sure modules [won't be reloaded before listeners][5],
- bumping the `xmpp` and `stun` dependencies in [rebar.config][6]/[mix.exs][7],
- [replacing][8] the `[warning]` logged by `ejabberd_stun` when multiple domains are configured and no auth realm is specified with a less alarming `[info]` message (as the realm is [irrelevant][9] for XEP-0215 authentication), and
- [adding][10] the module to the sample configuration file.

The module is shipped with [test cases][11], has already been tested by several admins, and is running in production on two of the servers I operate myself.

Many thanks to @ChaosKid42, @licaon-kter, @rom1dep and others for testing things and giving very useful feedback.

Closes #2947.

[1]: https://xmpp.org/extensions/xep-0215.html
[2]: https://github.com/jitsi/jitsi-meet/issues/5786
[3]: https://tools.ietf.org/html/draft-uberti-behave-turn-rest-00
[4]: https://github.com/coturn/coturn/wiki/turnserver#webrtc-usage
[5]: https://github.com/processone/ejabberd/pull/3235/commits/88f392721b254f18dd87d34e91416e7b19f7dd29
[6]: https://github.com/processone/ejabberd/pull/3235/commits/fc444ce5032fb1039742d9497e2e93f18b4126e7
[7]: https://github.com/processone/ejabberd/pull/3235/commits/d7c696f97cd137e8a5ebae5e6638e9542e744882
[8]: https://github.com/processone/ejabberd/pull/3235/commits/6edc8af16032d4ddad9ce5ec8675515281dd7ade
[9]: https://tools.ietf.org/html/draft-uberti-behave-turn-rest-00#section-4.2
[10]: https://github.com/weiss/ejabberd/blob/xep-0215/ejabberd.yml.example#L206
[11]: https://github.com/processone/ejabberd/pull/3235/commits/223eb01b3ea50424173aa3dc1dc7623a27c1477f


licaon-kter:
Running this for a week for testing Conversations.  👍

Neustradamus:
@weiss: Thanks for your good job!

The missing important part is IPv6, hope soon too :)

weiss:
This module can be used just fine with STUN/TURN servers that support IPv6.

E-404:
This module work with SCRAM passwords?

licaon-kter:
@E-404 No, it generates short time credentials.

PR #3246, opened 2020-04-30
Little changes

Neustradamus:
Good order: A->Z


Neustradamus:
@p1bot: I confirm that you have a problem, it is already done for me.
cc @mremond.

badlop:
I've applied the changes that produced a significantly better user experience. And I've retracted the changes that were just cosmetic or produced minimal benefits.

PR #3255, opened 2020-05-05
Increase default shaper values

weiss:
The `normal` shaper value of 1,000 B/s in the example configuration, which is applied to `ejabberd_c2s` connections, delays the negotiation of audio/video streams significantly.  Clients such as Movim or Conversations generate up to 15 kB of Jingle traffic to negotiate the streams, so it takes up to 15 seconds until the streams are established after a call is accepted.

On the servers I operate myself, I had bumped the `normal`/`fast` shapers to 50,000/500,000 B/s a long time ago (as the old defaults seemed too restrictive to me).  Therefore, we hadn't noticed this problem during our A/V tests.  This PR bumps the shaper values in the example configuration file to the slightly more conservative values 20,000 and 200,000.

### DIFF
--- a/ejabberd.yml.example
+++ b/ejabberd.yml.example
@@ -127,8 +127,8 @@ api_permissions:
       - connected_users_number
 
 shaper:
-  normal: 1000
-  fast: 50000
+  normal: 20000
+  fast: 200000
 
 shaper_rules:
   max_user_sessions: 10


licaon-kter:
Testing confirmed this for me too. Thanks

prefiks:
Hello,

In general i think increasing default rate is probably ok thing to do, 1k/s nowadays seems a bit small, with omemo/pubsub getting bigger we probably need that. 20k/s seems a bit high, when someone tries to abuse things, but i think we could get better behaviour for normal clients if we changed another parameter in shaper - it's burst size.

You can configure shaper in two ways:
```
shaper:
  normal: 1000
  not_normal:
    rate: 1000
    burst_size: 20000
```
(if you use first way you basically telling to use same value for burst_size as for rate).

With burst size you can allow client to still send more data, but it's amount still can be clamped reasonably (how this works is that each connection is allowed to send `burst_size` of data before processing is delayed, and that amount is replenished by `rate` each seconds but never to more than `burst_size` allows.

I think having normal shaper set to like 5k/s and 20k burst_size should work ok with that, client still should be able to send quite a bit of data at once, but still have limited amount of data to send on constant basis. Of course we can tweak those numbers further, and test to see what works ok with that scenario would be great, but i think we should tweak it that way instead of increasing rate without touching burst size.

weiss:
Totally makes sense.  I just wasn't aware of the existence of `burst_size` (or forgot about it) :smile:  

Sounds perfect, thanks; I'll do some further testing.

mremond:
I think the default value could still be “upgraded”: Maybe have a default burst_size of 20KB and a default shaper of 2 or 3 KB ?

prefiks:
Hello,

I'm gonna commit something like this:
```
shaper:
  normal:
    rate: 3000
    burst_size: 20000
  fast: 100000
```
(bump normal from 1k, to rate to 3k, and add 20k burst_size, and also increase fast that is used for s2s from 50k to 100k).

prefiks:
I commited that change above in 4580feaa3c7162b6161bbcbce44514a8da8bd90a

Issue #3270, opened 2020-05-14
[FreeTDS][SQL Server] Incorrect syntax near the keyword  'as'


yuriyz-w:
## Environment

- ejabberd version: 20.03
- Microsoft SQL Server 2016 (SP1) (KB3182545) - 13.0.4001.0 (X64)

## Errors from error.log/crash.log

Apr 30 10:18:31 221100002fa7 ejabberd[12314]: [error] <0.575.0>@ejabberd_sql:check_error:1286 SQL query 'select jid = '101@d-luk09.wildixin.com' as G, affiliation, subscriptions from  pubsub_state where nodeid=1
 and jid in ('101@d-luk09.wildixin.com', '101@d-luk09.wildixin.com/cticonnect-101-6370-0628-2488')' failed: <<"[FreeTDS][SQL Server]Incorrect syntax near the keyword 'as'.">>                                     
Apr 30 10:18:31 221100002fa7 ejabberd[12314]: [error] <0.575.0>@ejabberd_sql:check_error:1279 SQL query 'Q78217197' at {node_flat_sql,900} failed: <<"[FreeTDS][SQL Server]Incorrect syntax near the keyword 'as'."

## Bug description
SQL computed field syntax is not compatible with MSSQL

prefiks:
Should be fixed by merged pull request.

PR #3271, opened 2020-05-14
Make SQL query more generic for MSSQL compatibility

yuriyz-w:
This changes get rid of simple calculate field from node_flat_sql module
and closes issue #3270.
Used SQL syntax is not compatible with MSSQL.
Just do compare a field on ejabberd side (do not use different query for MSSQL.).


prefiks:
Hello,

I will merge this, i had this part to make this query return smaller amount of data, but if this doesn't work properly on mssql, then let use simpler query.

PR #3356, opened 2020-08-11
avoid preparing statements for pgsql

oxpa:
When using with pgbouncer or other Postgresql balancer ejabberd doesn't behave well as it uses anonymous prepared statements which are valid for a single connection only.

The proposed change removes prepared statements from ejabberd and instead relies on p1_pgsql to prepare and execute statement (see p1_pgsql/pull/5).

This change also requires new pq_pgsql version, but I can't change rebar config as my pull request to pgsql is not yet accepted (at the time of me writing this message).

I'm not 100% sure if this will work properly in all cases, but I'm willing to fix any mistakes I left behind.

### DIFF
--- a/src/ejabberd_sql.erl
+++ b/src/ejabberd_sql.erl
@@ -597,36 +597,7 @@ sql_query_internal(#sql_query{} = Query) ->
 		mssql ->
 		    mssql_sql_query(Query);
                 pgsql ->
-                    Key = {?PREPARE_KEY, Query#sql_query.hash},
-                    case get(Key) of
-                        undefined ->
-                            Host = State#state.host,
-                            PreparedStatements =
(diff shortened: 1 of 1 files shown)


licaon-kter:
So this doesn't work? https://docs.ejabberd.im/archive/20_04/toplevel/#sql-prepared-statements

oxpa:
Ouch!
I actually patched an earlier version of ejabberd which didn't have this option and then applied the patch to current version.
So it may happen the patch doesn't make sense anymore :-\

Neustradamus:
@oxpa: With latest ejabberd, your patch is not needed?
Have you tested without?

oxpa:
Yes, this patch is not needed on newer versions of ejabberd

PR #3381, opened 2020-09-04
Allow for filtering outgoing s2s stanzas

weiss:
Let `s2s_send_packet` hook callbacks filter/drop stanzas, analogous to [the `s2s_receive_packet` hook][1].

[1]: https://github.com/processone/ejabberd/blob/20.07/src/ejabberd_s2s_in.erl#L221-L226


prefiks:
Having both send and receive be able to filter seems sane, so let's merge this. Consumers would have to be modified, but i don't think there is that much of them.

PR #3391, opened 2020-09-17
Add 'gitonly_deps' list to rebar config/script

nosnilmot:
Add list of dependencies that should only be built from git, to support building with rebar3 where deps do not have hex packages (or where the package versions do not directly map to git tags).

This is required for elixir and luerl deps.


prefiks:
If they don't have hex packages, then indeed we need something like this for rebar3, so i will merge it.

PR #3392, opened 2020-09-17
rebar3 plugin to support configure-deps command

nosnilmot:
To allow running configure on dependencies prior to compilation, add a rebar3 plugin to support the 'configure-deps' command introduced for rebar2 in a7639fd4


prefiks:
Thanks!

PR #3422, opened 2020-11-07
web admin display garbled

cisiqo:
   description.

In the ejabberd web admin , other country language may dispaly garbled(eg. Japanese, Chinese ...), like the url `/admin/node/ejabberd@localhost/`, `/admin/node/ejabberd@localhost/db/`

Yes.

3. Do you provide tests? How can we check the behavior of the code?

I test it in my environment. In the web admin just change the chrome language it will appear.

### DIFF
--- a/src/str.erl
+++ b/src/str.erl
@@ -286,7 +286,7 @@ suffix(B1, B2) ->
 -spec format(io:format(), list()) -> binary().
 
 format(Format, Args) ->
-    unicode:characters_to_binary(io_lib:format(Format, Args)).
+    list_to_binary(io_lib:format(Format, Args)).
 
 
 -spec sha(iodata()) -> binary().


prefiks:
Hello,

So looks like problem here is with doing conversion to utf-8 of data from translate:translate that is already converted to utf-8. Unfortunately in other places we are using str:format on data that is not converted, so we can't just remove that unicode conversion here, it will broke other places.

What we really need to do is probably adding
```
str:translate_and_format(Lang, String, Args) ->
   str:format(unicode:characters_to_list(translate:translate(Lang, String)), Args).
```
and replace all instances of `str:format(translate:translate(...))` with that.

I could try to do that if you don't want to spend time trying to handle this that way.

prefiks:
I did implement what i described in comment above in commit 81e872c11066fec3cbc92506d437a9452cffb793. 

Thanks for reporting this issue, i didn't notice that, even if my language has some characters that could be affected by this.

PR #3429, opened 2020-11-19
Let systemd start ejabberd in foreground

mtdcr:
Daemons started by systemd shouldn't fork into the background if
possible, because if multiple forked processes exist, systemd has
a hard time determining the main process ID.

In a memory constrained environment, the OOM killer may cause
ejabberd to exit without any trace. Because epmd keeps running,
systemd wouldn't notice the error condition, and as a result it
won't restart the server.

With ejabberd running in foreground, systemd is able to obtain the
correct exit code (137 in this case, instead of 0) and schedules a
restart. The administrator can then see what happend by looking at
systemctl status ejabberd.

### DIFF
--- a/ejabberd.service.template
+++ b/ejabberd.service.template
@@ -3,13 +3,12 @@ Description=XMPP Server
 After=network.target
 
 [Service]
-Type=forking
 User=ejabberd
 Group=ejabberd
 LimitNOFILE=65536
 Restart=on-failure
 RestartSec=5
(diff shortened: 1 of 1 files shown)


weiss:
I think we now need a proper way to configure whether log messages should continue to be written to file(s) and/or stdout (and therefore to the systemd journal).  Right now, ejabberd will write to both.

prefiks:
@weiss 3516d2053cc8e0375310dc3be86529868fa196f1

weiss:
You're the best :smile:

(I guess real systemd fans might want the opposite: Just log to the journal.  But I can have a look myself.)

PR #3446, opened 2020-12-13
feat: add hook to check decoded JWT after success authentication

pouriya:
This change makes it easy to check our custom fields (e.g. [`aud`](https://tools.ietf.org/html/rfc7519#section-4.1.3)).

### DIFF
--- a/src/ejabberd_auth_jwt.erl
+++ b/src/ejabberd_auth_jwt.erl
@@ -102,10 +102,11 @@ check_jwt_token(User, Server, Token) ->
                                 error ->
                                     false;
                                 {ok, SJID} ->
-                                    try
-                                        JID = jid:decode(SJID),
-                                        (JID#jid.luser == User) andalso
-                                        (JID#jid.lserver == Server)
+                                    try jid:decode(SJID) of
+                                        JID ->
+                                            (JID#jid.luser == User) andalso
+                                            (JID#jid.lserver == Server) andalso
+                                            ejabberd_hooks:run_fold(check_decoded_jwt, Server, true, [Fields, Signature, User])
                                     catch error:{bad_jid, _} ->
                                             false
                                     end


prefiks:
Thanks, looks like fine additon.

pouriya:
NP. I also wanted to discuss something else. In JWT authentication the authentication server should put the entire jid/bare-jid into JWT. Sometimes there are other services (except Ejabberd) that can use this JWT too and XMPP jid means nothing to them! So they probably should cut the username part or do something else. But with a few changes in this code, many services can use the same JWT:
```erlang
case maps:find(JidField, Fields) of
    error ->
        false;
    {ok, ID} when is_integer(ID) ->
        erlang:integer_to_binary(ID) == User andalso 
        ejabberd_hooks:run_fold(check_decoded_jwt, Server, true, [Fields, Signature, User]);
    {ok, SJID} when is_binary(SJid)  ->
        try jid:decode(SJID) of
            JID ->
                (
                    (JID#jid.luser == <<>> andalso JID#jid.lserver == User) orelse 
                    (JID#jid.luser == User andalso JID#jid.lserver == Server)
                ) andalso
                ejabberd_hooks:run_fold(check_decoded_jwt, Server, true, [Fields, Signature, User])
        catch error:{bad_jid, _} ->
            false
       end;
    _ -> % {ok, _}
        false
end;
```

What is your opinion?

pouriya:
@prefiks

prefiks:
Don't really like this. You can use different field for different service if you have this need, and this starts to get too specialized, and it won't even handle all cases. Alternativelly if you see need for that we could add another hook, that could process Fields earlier.

pouriya:
@prefiks I see. So what about this:  
```erlang
start(Host) ->
    ejabberd_hooks:add(check_decoded_jwt, Host, ?MODULE, check_decoded_jwt, 100),
    case ejabberd_option:jwt_key(Host) of
	undefined ->
	    ?ERROR_MSG("Option jwt_key is not configured for ~ts: "
		       "JWT authentication won't work", [Host]);
	_ ->
	    ok
    end.

stop(_Host) ->
    ejabberd_hooks:delete(check_decoded_jwt, Host, ?MODULE, check_decoded_jwt, 100).

% ...

check_decoded_jwt(_Acc, _Fields, _Signature, Server, User) ->
    JidField = ejabberd_option:jwt_jid_field(Server),
    case maps:find(JidField, Fields) of 
        {ok, SJid} when is_binary(SJid) ->
            try
                JID = jid:decode(SJID),
                (JID#jid.luser == User) andalso (JID#jid.lserver == Server)
            catch error:{bad_jid, _} ->
                false
            end;
        _ -> % error | {ok, _UnknownType}
            false
    end.

% ...

check_jwt_token(User, Server, Token) ->
    JWK = ejabberd_option:jwt_key(Server),
    JidField = ejabberd_option:jwt_jid_field(Server),
    try jose_jwt:verify(JWK, Token) of
        {true, {jose_jwt, Fields}, Signature} ->
            ?DEBUG("jwt verify: ~p - ~p~n", [Fields, Signature]),
	    case maps:find(<<"exp">>, Fields) of
                error ->
		    %% No expiry in token => We consider token invalid:
		    false;
                {ok, Exp} ->
                    Now = erlang:system_time(second),
                    if
                        Exp > Now ->
                            ejabberd_hooks:run_fold(check_decoded_jwt, Server, true, [Fields, Signature, Server, User])
    % ...
```

pouriya:
@prefiks Also a note about debugging, We can put something about time difference in `?DEBUG` macro:
```erlang
check_jwt_token(User, Server, Token) ->
    JWK = ejabberd_option:jwt_key(Server),
    JidField = ejabberd_option:jwt_jid_field(Server),
    try jose_jwt:verify(JWK, Token) of
        {true, {jose_jwt, Fields}, Signature} ->
            Now = erlang:system_time(second),
            % Ofc that log has its time, but to check that JWT was expired or not 
            % we have to check it's time manually.
            % But here we put system time in timestamp format in log and 
            % we can simply see/check time differenece just by reading logs:
            ?DEBUG("verifying JWT at system time ~p: ~p - ~p~n", [Now, Fields, Signature]),
	    case maps:find(<<"exp">>, Fields) of
                error ->
		    %% No expiry in token => We consider token invalid:
		    false;
                {ok, Exp} ->
                    if
                        Exp > Now -> 
```

prefiks:
I was more thinking about
```
Exp > Now ->
  Fields2 = ejabberd_hooks:run_fold(process_jwt_fields, Server, Fields, [Server]),
  case maps:find(JidField, Fields2) of
  ...
```

pouriya:
What I say is more like `mod_client_state`. It adds a callback hook for `csi_activity` [here](https://github.com/pouriya/ejabberd/blob/673842d7a380ebf26c42b2bd702aee04658f9575/src/mod_client_state.erl#L200) and then it calls its [here](https://github.com/pouriya/ejabberd/blob/673842d7a380ebf26c42b2bd702aee04658f9575/src/mod_client_state.erl#L235). So this module has a default behaviour for that event (`csi_activity`) and also others can use this hook for their own purpose.

pouriya:
@prefiks BTW if you are more convinced with what you said, It's ok and this can get the job done. Also if you are convinced with the other change (adding `Now` to `?DEBUG` macro) tell me.

pouriya:
@prefiks ping

prefiks:
Now that i took a second look, i am ok with your proposed code (that one that moves verification of jid to check_decoded_jwt hook), i am also fine with extending debug data. If you are fine with preparing new merge request with that change then go ahead, if not i will probably try to do it as some point.

pouriya:
Thank you. I will send a new PR as soon as possible.

pouriya:
@prefiks I opened a new PR #3460.

PR #3448, opened 2020-12-15
Reword auth_scram_hash info

licaon-kter:


prefiks:
Thanks, looks much better.

PR #3452, opened 2020-12-18
Fix sqlite3 dependency version

nosnilmot:
erlang-sqlite3 versions got messed up, causing ejabberd build to regress to 1.1.6 from Jan 2018 when using rebar3 which fails to build with OTP23.
Update to correctly tagged version 1.1.9, which also has matching hex.pm package, fixes this.


prefiks:
Thanks, scripr that i use to update deps was thinking that "v1.1.6" is newer than "1.1.9", and thats why we got this version.

nosnilmot:
I think at the time you updated the deps, the tag "v1.1.6" was the newest ("1.1.9" was tagged later the same day), but "1.1.6" is nearly 3 years old and that is what actually got referenced somehow. https://github.com/processone/erlang-sqlite3/tags is an interesting progression of numbers, it seems to be haunted :-)

PR #3471, opened 2021-01-06
Integrate nicely with systemd

weiss:
Support [systemd's watchdog feature][1] and enable it by default in the unit file, so that ejabberd is auto-restarted if the VM becomes unresponsive.  Also, set the systemd [startup type][2] to `notify`, so that startup of followup units is delayed until ejabberd signals readiness.  While at it, also notify systemd of [configuration reload][3] and [shutdown][4] states.

[1]: http://0pointer.de/blog/projects/watchdog.html
[2]: https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=
[3]: https://www.freedesktop.org/software/systemd/man/sd_notify.html#RELOADING=1
[4]: https://www.freedesktop.org/software/systemd/man/sd_notify.html#STOPPING=1

### DIFF
--- a/ejabberd.service.template
+++ b/ejabberd.service.template
@@ -3,14 +3,17 @@ Description=XMPP Server
 After=network.target
 
 [Service]
+Type=notify
 User=ejabberd
 Group=ejabberd
 LimitNOFILE=65536
 Restart=on-failure
 RestartSec=5
+WatchdogSec=30
 ExecStart=@ctlscriptpath@/ejabberdctl foreground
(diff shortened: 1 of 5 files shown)


prefiks:
Hi,

That looks pretty nice. I was not sure that we started to listen to sockets when `ejabberd_systemd:ready()` was called but i think proc_lib:init_ack() that listeners calls should make us wait long enough when starting ejabberd_sup tree.

I think you can go ahead and merge it, unless you plan to do some extra work on this.

weiss:
> i think `proc_lib:init_ack()` that listeners calls should make us wait long enough when starting ejabberd_sup tree.

Yes, exactly, the listener supervisor startup should block until all listeners called `proc_lib:init_ack()`.

Many thanks for your review.

PR #3478, opened 2021-01-12
XMPP + ejabberd

Neustradamus:
Replace:
- Jabber to XMPP
- Ejabberd to ejabberd

### DIFF
--- a/man/ejabberd.yml.5
+++ b/man/ejabberd.yml.5
@@ -382,7 +382,7 @@ Git repository\&. Please refer to that module\(cqs README file for details\&.
 \fBauth_password_format\fR: \fIplain | scram\fR
 .RS 4
 The option defines in what format the users passwords are stored\&.
-\fIplain\fR: The password is stored as plain text in the database\&. This is risky because the passwords can be read if your database gets compromised\&. This is the default value\&. This format allows clients to authenticate using: the old Jabber Non\-SASL (XEP\-0078), SASL PLAIN, SASL DIGEST\-MD5, and SASL SCRAM\-SHA\-1\&.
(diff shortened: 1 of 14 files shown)


badlop:
This was already submitted in https://github.com/processone/ejabberd/pull/3469, reviewed, and some specific changes were rejected. Why are they submitted again?

Neustradamus:
@badlop: I know that win32 folder is very old.
You can remove no?
Same for transport files?

Why you do not accept changes in other files?

badlop:
> You can remove no?

Probably yes, but I am not 100% sure. So, I've filed https://github.com/processone/ejabberd/issues/3487 so that can be discussed.

> Why you do not accept changes in other files?

XEP-0078 is related to Jabber, so that must be mentioned.

"Outdated Jabber bots" are "Outdated Jabber bots". They are not "Outdated XMPP bots".

Websocket and BOSH informative pages: as those pages are displayed to end users that may not know "XMPP" term, both terms should be displayed, Jabber and XMPP

Neustradamus:
@badlop:
-> Jabber is the old name of XMPP
- https://www.jabber.org/
- https://en.wikipedia.org/wiki/XMPP

-> Jabber bots = XMPP bots

BOSH are XEPs, XMPP Extensions:
- https://xmpp.org/extensions/xep-0124.html
- https://xmpp.org/extensions/xep-0206.html

WebSocket is an RFC:
- https://tools.ietf.org/html/rfc6455

XMPP and WebSocket is an RFC:
- https://tools.ietf.org/html/rfc7395

Neustradamus:
@badlop: Can you reopen my PR?

badlop:
Can you stop lobbying?

Neustradamus:
@badlop: But it is to be only perfect :)
It is better to reopen than a new PR...

PR #3479, opened 2021-01-13
http -> https

Neustradamus:
Replace:
- http -> https

Drop www. from
- ejabberd.im links
- erlang.org links


badlop:
```
-		[{<<"xmlns">>, <<"http://www.w3.org/1999/xhtml">>},
+		[{<<"xmlns">>, <<"https://www.w3.org/1999/xhtml">>},
```

I don't think this is a good idea. The namespace is http://www.w3.org/1999/xhtml not what you say

badlop:
Also, what's the point in updating all URLs for HTTPS alternatives? If the website and the client support it, they update it automatically, if not, they at leeast can use the non-encrypted one.

Neustradamus:
@badlop: Search on GitHub, the number of pages wirh https instead of http for w3 links

badlop:
@Neustradamus Read the specs.

And check https://www.w3.org/1999/xhtml/ source code. Even W3C itself has:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-stylesheet type="text/css" href="http://www.w3.org/StyleSheets/TR/base"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN"
      "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-1.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
```

And search in github code:
```
 http://www.w3.org/1999/xhtml/ results: 59,000,000
https://www.w3.org/1999/xhtml/ results:    128,000
```

Neustradamus:
@badlop: I have updated this PR without http://www.w3.org changes.

src/ejabberd_piefxis.erl
```diff
@@ -55,7 +55,7 @@
 %%%==================================
 %%%% Import file
 -define(NS_PIE, <<"urn:xmpp:pie:0">>).
--define(NS_PIEFXIS, <<"http://www.xmpp.org/extensions/xep-0227.html#ns">>).
+-define(NS_PIEFXIS, <<"https://xmpp.org/extensions/xep-0227.html#ns">>).
```
badlop, line 58: No, the namespace is http://www.xmpp.org/extensions/xep-0227.html#ns and you can't change it just because you want to.

src/ejabberd_oauth.erl
```diff
@@ -521,7 +521,7 @@ process(_Handlers,
     Top =
         ?DIV(<<"section">>,
              [?DIV(<<"block">>,
-                   [?A(<<"https://www.ejabberd.im">>,
+                   [?A(<<"https://ejabberd.im/">>,
```
badlop, line 524: No, because all those https://ejabberd.im/ links redirect to https://www.ejabberd.im/

Neustradamus:
@badlop: Ready?

Neustradamus:
@badlop: You have forgotten in 21.01...

Neustradamus:
@badlop: I need to do a new PR?
I have solved your remarks.

Neustradamus:
@badlop: It is good now or there is always a problem?

badlop:
There were several mistakes in the initial commit, just by looking at them. You've fixed some, but I am not confident there may be more errors. It would require a lot of time to verify all them.

On the other hand, this massive change is purely cosmetic, it is not an improvement, so the time and effort investment doesn't pay.

And there are 163 issues with real bugs and real improvements, so let's invest time and effort in them.

PR #3486, opened 2021-01-14
ejabberd_oauth language reworked

comradekingu:


src/ejabberd_oauth.erl
```diff
@@ -547,7 +547,7 @@ process(_Handlers,
                     ?C(<<" is maintained by ">>),
                     ?XAC(<<"a">>,
                          [{<<"href">>, <<"https://www.process-one.net">>},
-                          {<<"title">>, <<"ProcessOne - Leader in Instant Messaging and Push Solutions">>}],
+                          {<<"title">>, <<"ProcessOne—Leader in Instant Messaging and Push Solutions">>}],
```
comradekingu, line 550:
```suggestion
                          {<<"title">>, <<"ProcessOne—Instant Messaging and Push Solutions">>}],
```

badlop: Your next step should be something like
```diff
- 2001 - A Space Oddyssey
+ 2001 - A Guy Takes a Peaceful Walk on Sky
```

badlop:
Everything else looks perfect. Thanks for reporting! It will also be updated in the documentation page https://docs.ejabberd.im/developer/ejabberd-api/admin-api/

src/ejabberd_oauth.erl
Neustradamus: I have discovered the problem here.
Better to revert this part.

comradekingu: The problem being?

Neustradamus: Original:
- "ProcessOne - Leader in Instant Messaging and Push Solutions"

Your version:
- "ProcessOne—Leader in Instant Messaging and Push Solutions"

comradekingu: That is an em dash. I meant to do that, if it makes a difference.

PR #3488, opened 2021-01-16
mod_roster language reworked

comradekingu:


badlop:
Thanks! I'll make sure the changes get applied to manual and documenation https://github.com/processone/ejabberd/issues/3490

badlop:
By the way, @comradekingu, can you please sign the CLA mentioned earlier?

comradekingu:
@badlop I don't sign CLAs on account of wanting to answer "no" to ever having signed any, should a future employer ask.
Moreover this isn't copyrightable, and thus not subject to licensing.
If you want to copy it over to your own branch, go ahead, these are merely suggestions.
When I click the link I don't get to read what I am signing before logging in. CLAs usually just undermine licensing.
If you want fewer issues in that department, get a cure-clause for GPLv2, or upgrade to v3+. How it isn't AGPLv3+ for what it is, I don't know, but then again I am just a user.

Edit: The translation license in https://hosted.weblate.org/projects/ejabberd/#information is GPLv2-only, which seems problematic. No CLA to be had on there though, which is good.

comradekingu:
@badlop

About the new https://hosted.weblate.org/contributor-agreement/ejabberd/ejabberd/?next=/projects/ejabberd/ejabberd/

> The ejabberd project is released under the GNU GPL v2.0 license, and the copyright holder is ProcessOne. In order to accept your contribution, if you haven't done so before, we need you to sign the ProcessOne Contribution License Agreement as either a corporation or an individual. If you cannot accept the terms laid out in the agreement, unfortunately, we cannot accept your contribution.

The license part is already specified and agreed to under 6.4 of https://hosted.weblate.org/legal/terms/
If you don't read the required parts of licensing for services you use yourselves, how do you expect anyone else to read your extra terms and conditions?

For some reason the translation license is "GNU General Public License v2.0 only". https://hosted.weblate.org/projects/ejabberd/ejabberd/#information Change it to + and hope nobody complains.

>This is part of the legal framework of the open-source ecosystem that adds some red tape, but protects both the contributor and the company / foundation behind the project. It also gives us the option to relicense the code with a more permissive license in the future.

This is not part of the ecosystem framework at all, it is part of the non-libre software framework, and it permits licensing the code as such.

> So, please enter this page and login with your Github account to read the agreement. Finally, you can sign it electronically: https://cla.process-one.net/

Still no thanks, scrap the nb_NO translation if you want.

Change your CLA to require full and strong copyleft for any license changes if you want to not trick people.

badlop:
Thanks for your interest and suggestions. I'll revert the corresponding contributions before release.

weiss:
> CLAs usually just undermine licensing.

PR comments are probably not the best place to discuss such topics.  However, as a contributor who signed the CLA, let me mention that I'm fine with how the non-exclusive copyright transfer this CLA asks for allows for dual-licensing.  If this helps with a business model that involves publishing open source software, I'm all for it, and I'd appreciate if more businesses would go for it.  The alternative you suggest seems to be GPL-only licensing; while in reality, the alternative might be that the software in question simply wouldn't exist, or only as proprietary code.

comradekingu:
@weiss It doesn't only allow dual-licensing, it allows monolicensing into any license, including non-libre licences.
That isn't a business model involving "open source" at all.

Whatever the business model of doing the current license is built on, it works with the current license.
The alternative I propose is the freedom to change license into anything that is full and strong copyleft, not just GPL, ensuring FSF doesn't have any power over even dictating the direction with GPLv4.
The _opposite_ would be something like licensing all translations GPLv2-only, which is something this project actually does (for no good reason AFAICT).


Networking software that isn't AGPL already makes the distribution clause in GPL pointless btw.

ejabberd is the type of project to waste everyone's time on reverting commits found to be legitimate, and scrap entire translations.

Signing away all your rights isn't a formality, nor is losing users of 15+ years (which you just did).
Guess how good seeding that sentiment into the ecosystem is for business.

weiss:
So in your view, copyleft software should never be cross-financed by a proprietary version of that software, regardless of business constraints.  This is where we disagree, and as far as I'm concerned we can just agree to disagree.

comradekingu:
@weiss 

As for sound business decisions, this isn't it.
This isn't functionally copyleft software either, it only masquerades as it by not having the CLA in the root directory. Underhanded.

To actually deal with the issue, the CLA not only undermines what gives copyleft software the protection it has (shared copyright), but makes the issue of being able to re-license into being able to close off any further development of the software in libre form. In my view copyleft software should never be made proprietary. As far as distribution goes, I think sharing it over a network is a caveat in that respect, which is why I like the network clause, regardless of license.

I certainly don't think it should ever happen through "gives us the option to relicense the code with a more permissive license in the future.", nor is it what said statement argues. Underhanded?

If you think a business model can't survive sticking to the copyleft nature of a copyleft product by guaranteeing as much, it isn't copyleft. You will have a hard time arguing why it should get volunteer contributions under such a license.

Where we can agree to _fundamentally_ disagree is whether communication people rely on should ever have a built-in proprietary switch. I am all ears as to why there should be one.

To deal in the perfectly superfluous, why isn't there any mention of the CoC on Weblate?

weiss:
Seems you believe the difference in our views is due to you understanding how copyleft works, how that's affected by CLAs, how business works, and me missing those insights.  As I mentioned above, I think our disagreement is elsewhere, but I have no interest in continuing this discussion on GitHub.  Anyway, thanks for sharing your views.

comradekingu:
I don't know what insights you have missed. You got "in your view" about mine wrong, and based "this is where we disagree" on that.

comradekingu:
After a wave of censorship that sent user-base and donations for alternative platforms and systems through the roof there is no mention of it or donation effort to be found (?)

Hypothetically company uses other libraries and operating systems to do so, that are all libre software.
Notwithstanding, company reserves its own right to do what it will with its own code. I don't have a say here.

Pretending that it is copyleft also as a function of being copyleft for contributors, why? Why can't it put the CLA in the root directory like similar companies.

A further why is "non-exclusive" meaning every conceivable notion of copyright transfer as it relates to copyright protection?

Shouldn't people contributing be ensured the parts they are entitled to stay copyright for the parts it concerns?

I don't see a benefit to undermining that part. And don't give me the "you get to keep the parts". That isn't copyleft.

At least put company mouth where money isn't?

If you will, imagine the conceptual later-added CLA is gone from the translations, and they are lost due to making them MIT. Will this company just use GPLv2 content without a cure clause and hope that flies?

weiss:
I guess we both did a good job at explaining our points of view.  As there's no need for us to reach a consensus (you can just quit contributing, while I can continue), let's finally leave it at that.

PR #3506, opened 2021-01-29
Remove queryid from MAM fin element

paulfariello:
According to xep, fin element is not supposed to have a queryid attribute.


Neustradamus:
@paulfariello: Can you sign the CLA?
- You must validate the email...

weiss:
Thanks!  This should only be done for `mam:1` and `mam:2`.  With `mam:0`, the `<fin/>` element requires the `queryid` attribute.

PR #3507, opened 2021-01-29
Allow ejabberd to be compatible as a dependency for an Erlang project…

slezakattack:
… using rebar3

When I set ejabberd as a dependency in my Erlang project using rebar3, the "ejabberd.app.src.in" was breaking the build due to a tight dependency of running "configure" script before building. I've made a change to refactor this so that "configure" script can still be used as well as building ejabberd as a dependency for a rebar3-enabled Erlang project.


badlop:
Travis reported a false problem. It's unrelated to this PR. When this commit is rebased to recent master, test passes correctly.

Anyway, I've also checked manually that ejabberd compiles and starts correctly both using rebar2 and rebar3.

nosnilmot:
Replacing the test for `config.status` with a simple check for existence of `configure` will result in unnecessary invocations of `configure` in normal builds.

`vars.config` still needs to be created for accurate build, which `configure` would normally do, but this PR ignores that.

I faced a similar problem of wanting to include ejabberd as a dependency in another build, so I used something similar to the [`configure_deps` ](https://github.com/processone/ejabberd/tree/master/_checkouts/configure_deps) plugin to run `configure` as required on dependencies (ejabberd and others).

slezakattack:
> Replacing the test for `config.status` with a simple check for existence of `configure` will result in unnecessary invocations of `configure` in normal builds.
> 
> `vars.config` still needs to be created for accurate build, which `configure` would normally do, but this PR ignores that.
> 
> I faced a similar problem of wanting to include ejabberd as a dependency in another build, so I used something similar to the [`configure_deps` ](https://github.com/processone/ejabberd/tree/master/_checkouts/configure_deps) plugin to run `configure` as required on dependencies (ejabberd and others).

Thanks for calling this out. After some testing, it seems restoring the check for `config.status` still works for my needs. I will make that change.

badlop:
I noticed that some _included_applications_ from _ejabberd.app.src.in_ are not mentioned now: fast_tls, fast_yaml, p1_utils.
And there are some new added in _ejabberd.app.src.script_: jiffy, mqtree, p1_acme, p1_oauth2, yconf.

slezakattack:
Sorry, I didn't catch that the included_application list has changed. I've made the update.

slezakattack:
Hi, it's been about a month since this PR was submitted. I wanted to follow up and see if there's anything else required of this PR in order to move it forward. Thanks!

badlop:
I have a pair of small cosmetic changes, but I guess they can be done after merging.

badlop:
Your PR adds this. Is it used anywhere?
```diff
+{package_name, "@PACKAGE_NAME@"}.
````

slezakattack:
Oh, it looks like "@PACKAGE_NAME@" was originally set in the "description" field in the ejabber.app.src.in file. Apologies, I put everything in the vars.config file first and must've forgot to remove it.

PR #3510, opened 2021-01-31
Add tempsub feature for pubsub nodes

nbruning:
This PR is a rewrite of EJAB-1659, which was written for Ejabberd 14 originally.

It provides an implementation for the "tempsub" option on pubsub node configurations. It makes all subscriptions be temporary: they are cleared whenever the session is removed.

- If you make a pubsub subscription on your *full jid*, the subscription is cleared when your resource goes offline
- If you make a pubsub subscription on your *bare jid* , the subscription is cleared when *any* of your resources go offline - the spec doesn't say anything about this scenario, and I couldn't find another implementation to borrow from)

As a consequence, multiple sessions (resources) do not clear each other's subscriptions, *if the subscription was made using a full jid*.

### DIFF
--- a/src/mod_pubsub.erl
+++ b/src/mod_pubsub.erl
@@ -57,7 +57,7 @@
     disco_local_identity/5, disco_local_features/5,
     disco_local_items/5, disco_sm_identity/5,
     disco_sm_features/5, disco_sm_items/5,
-    c2s_handle_info/2]).
+    c2s_handle_info/2, sm_open_connection/3, sm_remove_connection/3]).
 
 %% exported iq handlers
 -export([iq_sm/1, process_disco_info/1, process_disco_items/1,
@@ -318,6 +318,10 @@ init([ServerHost|_]) ->
 	?MODULE, remove_user, 50),
     ejabberd_hooks:add(c2s_handle_info, ServerHost,
 	?MODULE, c2s_handle_info, 50),
+    ejabberd_hooks:add(sm_register_connection_hook, ServerHost,
+	?MODULE, sm_open_connection, 50),
+    ejabberd_hooks:add(sm_remove_connection_hook, ServerHost,
+	?MODULE, sm_remove_connection, 50),
     case lists:member(?PEPNODE, AllPlugins) of
 	true ->
 	    ejabberd_hooks:add(caps_add, ServerHost,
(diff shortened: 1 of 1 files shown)


prefiks:
Hello,

While i am not opposing adding that feature, i don't particularly like how that is implemented, fetching pubsub nodes twice in session lifetime adds quite a bit to database load on busier servers, and we already and this can be bottleneck already. I think we can do better than that. I think we can store info about subscribed tempsub nodes in c2s state (after all we only need that info when session is alive), but it will require some refactoring.

We should be able to return info from `subscribe_node()` to tell that subscribed node was `tempsub` (`transaction()` call inside of it returns `#pubsub_node{}` TNode - we can extract options from it and check if it have this attribute set), but we would need to add ability to return more info from that, probably we could add additional return case {result, pubsub_response, metadata_about_tempsub_subscription}, and bubble up handling for that in it's caller (that is process_pubsub, iq_sm), and add info about subscription to returned iq result (by using `xmpp:set_meta()` on that. Then we should be extend `c2s_handle_info` to recognize `{route, #iq{meta = #{tempsub := SubData}}}` and add info about that SubData to c2s state. Then finally we will need to add code to `on_user_offline()` that will get temp nodes from c2s state, and call unsubscribe on them.

With that we will not add any cost for clients that aren't using tempsub, and i think it will be also more correct (looking at spec we should do unsubscribe only when session that subscribes ends, not any other concurent session).

weiss:
> I think we can store info about subscribed tempsub nodes in c2s state

This won't work for remote subscribers, though.

I agree this shouldn't be merged as-is.  I think the underlying problem is that we have no adequate presence-based subscription mechanism.  If we did, tacking that `pubsub#tempsub` feature on top would probably be easy (though I'm not sure I like it, and I'm unclear about some details, that full vs. bare JID subscription thing being just one of them).

For PEP nodes, our PubSub/CAPS code stores subscription info within the _publishers_ c2s state (duplicated into `n` sessions if the publisher is online with `n` clients).  It was designed this way under the assumption that PEP is used for stuff such as `mood` or `tune`, which is only interesting as long as the publisher is online.  These days, PEP is used for things such as avatars or distribution of public E2EE key material, which the subscriber wants to access regardless of the presence state of the publisher.  These use cases fail with our PEP code, and that failure is only somewhat hidden by a combination of (1) clients (such as Conversations) working around the problem (by explicitly querying the PEP nodes of offline contacts), and (2) publishers often being online 24/7 with their (Android) phones anyway.

To fix this properly, I think we should get rid of (ab)using the client state altogether.  Instead, I think `mod_pubsub` should maintain the list of presence-based subscriptions just like it maintains explicit subscriptions, except that presence-based subscriptions _could_ go into some memory table.  This would avoid a lot of ugliness and special-casing of PEP in our PubSub code, it would avoid duplicating subscription info across multiple sessions, and most importantly, it would fix the PEP behavior for offline publishers.  And it would probably make it trivial to implement that `pubsub#tempsub` thing by enforcing subscriptions to be stored within the presence subcribers table, if we really want that feature.

Neustradamus:
@nbruning: Can you look @prefiks and @weiss comments?
What do you think?

nbruning:
> @nbruning: Can you look @prefiks and @weiss comments?
> What do you think?

Hi,

I think their comments make sense, it's definitely not optimized code. But the refactors which are coined will take significant development effort, and I'm afraid I don't have the time to do that.

I can't estimate the potential performance problems of merging this as-is. If you decide not to go with my approach, I guess we should just close the MR until somebody is able to put more effort into this.

PR #3513, opened 2021-02-01
Fix specified key was too long

divine:
This adds the correct engine and sets it to utf8mb4 collate.

Prevents "Specified key was too long; max key length is 1000 bytes" from happening.


prefiks:
Hello,

Could you do similar change to mysql.sql (i think also only those two tables are missing that). Just to be consistent between those two.

divine:
@prefiks Hello,

Sure, you're right, I have added a new commit.

Please let me know if everything is ok.

Thanks!

prefiks:
Looks fine, so let's merge it.

Thank you for your contribution.

Issue #3527, opened 2021-02-12
LDAP TLS verify doesn't work


fpiecka:
## Environment

- ejabberd version: 21.01
- Erlang version: Erlang (SMP,ASYNC_THREADS,HIPE) (BEAM) emulator version 10.3.4
- OS: Linux (Debian Buster)
- Installed from: official deb

## Configuration:

```diff
--- /opt/ejabberd/conf/ejabberd.yml.orig	2021-02-12 15:35:31.438847080 +0100
+++ /opt/ejabberd/conf/ejabberd.yml	2021-02-12 15:36:09.802412442 +0100
@@ -17,7 +17,7 @@
 hosts:
   - "jabber.example.com"
 
-loglevel: 4
+loglevel: 5
 log_rotate_size: 10485760
 log_rotate_date: ""
 log_rotate_count: 1
@@ -129,6 +129,14 @@
       - status
       - connected_users_number
 
+
+auth_method: [ldap]
+ldap_servers: [ldap]
+ldap_encrypt: tls
+ldap_base: "dc=example,dc=com"
+ldap_filter: "(objectClass=shadowAccount)"
+ldap_tls_cacertfile: /var/tmp/ca.crt
+ldap_tls_verify: hard
 shaper:
   normal: 1000
   fast: 50000
```

## Errors from error.log/crash.log

```
2021-02-12 15:39:53.847 [debug] <0.560.0>@eldap:connect_bind:1051 Connecting to LDAP server at ldap:636 with options [{packet,asn1},{active,true},{keepalive,true},binary,{verify,2},{cacertfile,<<"/var/tmp/ca.crt">>}]
2021-02-12 15:39:53.858 [error] <0.560.0>@eldap:connect_bind:1073 LDAP connection to ldap:636 failed: received CLIENT ALERT: Fatal - Handshake Failure - {bad_cert,hostname_check_failed}
2021-02-12 15:39:53.858 [info] <0.727.0> TLS client: In state certify at ssl_handshake.erl:1372 generated CLIENT ALERT: Fatal - Handshake Failure - {bad_cert,hostname_check_failed}
```

## Bug description

When I set ```ldap_tls_verify: hard```, ejabberd won't connect to LDAP.

This worked correctly in ejabberd 18.04.

Ejabberd 18.06 and later versions don't work.

It doesn't work with our OpenLDAP server and it doesn't work with a simulated TLS server using socat. Both are properly validated when checked with openssl s_client.

[...]

PR #3528, opened 2021-02-15
fix eldap certificate verification

nosnilmot:
Reported in #3527. Add hostname matching function, and specify SNI

Also, OTP 23 dropped backwards compatibility for 0, 1, 2 values for verify, so replace with combination of verify_none/verify_peer and fail_if_no_peer_cert as appropriate


prefiks:
Looks good, thank you.

PR #3533, opened 2021-02-18
Add question/answer-based CAPTCHA script

weiss:
This PR is submitted on behalf of Adrien Bourmault.  His commit adds a question/answer-based CAPTCHA script which is supposed to be more effective than our example script. 
 Example challenge: "Add 42 + 23."

I think there's two things to solve with the new script:

- It currently relies on `/bin/bash`.
- It currently only speaks English.

### DIFF
--- a/tools/captcha-ng.sh
+++ b/tools/captcha-ng.sh
@@ -24,13 +24,13 @@ INPUT=$1
 TRANSFORMATIONS=(INTRUDER SUM)
 DIGIT=(zero one two three four five six seven eight nine ten)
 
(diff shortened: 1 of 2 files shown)


prefiks:
Let's merge it, we can thing about improving it when it's in tree.

member7me:
CAPTCHA realisation is't effective :( It's a pity :( Today received 2500 new spam registration.
XMPP and ejabberd are slowly killed by spam.
Should close registration, as last 2 years :(

licaon-kter:
@member7me Which script did you use? 

No other protection?

PR #3554, opened 2021-03-21
add some metrics to the "hooks_type_test" module

fdie:
I found "hooks_type_test" very usefull and added some metrics to evaluate the performance of the various modules on a runnning system.
I don't know all the usages of "hooks_type_test", but here is how I use it, maybe helpfull for others ?
I generally grab some hook call with **redbug** :

```erl
erl> redbug:start("mod_mam:user_send_packet").
```

then define the args of the trace

```erl
erl> Acc = {{message,...},#{...}}.  % copy paste the args of the call
```

To feed in the associated hook method in **hooks_type_test** to get some execution timings:

```erl
erl> hooks_type_test:user_send_packet(Acc).
{
   {{message,...}, #{...}},
   % added map with detailed metrics and totals
   #{
      measures => [
       "{microseconds, reductions, {module, function}}",
        {2,14,{mod_privacy,user_send_packet}},
        {1,12,{mod_caps,user_send_packet}},
        {34,335,{mod_mam,user_send_packet}},
        {12,68,{mod_mam,user_send_packet_strip_tag}},
        {1,13,{mod_vcard_xupdate,user_send_packet}},
        {error,14,{mod_service_log,log_user_send}},
        {1,12,{mod_privilege,process_presence_out}},
        {44,28,{mod_ping,user_send}},
        {1,12,{mod_multicast,user_send_packet}},
        {error,12,{mod_metrics,user_send_packet}},
        {121,266,{mod_carboncopy,user_send_packet}}],
      total_microseconds => 217,
      total_reductions => 786
   }
}
```

I may be wrong, but I don't think this module is used in some specific context, 
thus it should be possible to modifiy the return values without breaking anything.
If I'm wrong the durations may be rather sent to the console to restore the former return value.

I kept the previous version with all the explicit method calls as commented out code.

Here is a sample of the generated code:

```erlang
%% called at ejabberd_c2s.erl:923
roster_get(Acc0, A) ->
    %% Acc1 = mod_roster:get_user_roster(Acc0, A),
    %% Acc2 = mod_shared_roster_ldap:get_user_roster(Acc1, A),
    %% Acc3 = mod_shared_roster:get_user_roster(Acc2, A),
    %% Acc3.
    run_metrics([Acc0, A], [
          {mod_roster, get_user_roster}
         ,{mod_shared_roster_ldap, get_user_roster}
         ,{mod_shared_roster, get_user_roster}
    ]).
```

Always run metrics twice, sometimes you may measure without/with caching ;-)

### DIFF
--- a/tools/hook_deps.sh
+++ b/tools/hook_deps.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env escript
 %% -*- erlang -*-
 
+-define(ENABLE_METRICS, true).
+
 -record(state, {run_hooks = #{},
 		run_fold_hooks = #{},
 		hooked_funs = {#{}, #{}},
@@ -325,6 +327,7 @@ emit_module(RunDeps, RunFoldDeps, Module) ->
 	emit_export(Fd, RunFoldDeps, "run_fold hooks"),
 	emit_run_hooks(Fd, RunDeps),
 	emit_run_fold_hooks(Fd, RunFoldDeps),
+	emit_metrics_helper(Fd, ?ENABLE_METRICS),
 	file:close(Fd),
 	log("Module written to ~s~n", [File])
     catch _:{badmatch, {error, Reason}} ->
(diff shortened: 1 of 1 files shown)


Neustradamus:
@prefix: What do you think?

Neustradamus:
@prefiks: What do you think?

badlop:
@fdie , did you see https://github.com/processone/ejabberd/pull/3756 ?

Neustradamus:
@fdie: Have you seen the @badlop comment?

Neustradamus:
@fdie: Have you looked previous comments?

Neustradamus:
@fdie: Have you seen the @badlop comment?

PR #3557, opened 2021-03-25
fix gen_pubsub_node:get_state return value

fdie:
"get_state" is still a "gen_pubsub_node" callback that does not return a "result" tuple.

-callback get_state(NodeIdx :: nodeIdx(),
	Key :: ljid()) ->
    pubsubState().

3 ways to fix this :

- modify the call back to return {result, pubsubState()}
- accept a #pubsub_state{} as return value              <===== My fix
- accept any return values since errors are matched by "error" and "{error, _}"


fdie:
wrong git id

Neustradamus:
Good PR: https://github.com/processone/ejabberd/pull/3558

PR #3558, opened 2021-03-25
fix gen_pubsub_node:get_state return value

fdie:
"get_state" is still a "gen_pubsub_node" callback that does not return a "result" tuple.

```
-callback get_state(NodeIdx :: nodeIdx(),
Key :: ljid()) ->
pubsubState().
```

3 ways to fix this :

modify the call back to return {result, pubsubState()}
accept a #pubsub_state{} as return value <===== My fix
accept any return values since errors are matched by "error" and "{error, _}"

### DIFF
--- a/src/mod_pubsub.erl
+++ b/src/mod_pubsub.erl
@@ -705,8 +705,7 @@ remove_user(User, Server) ->
 					case node_action(Host, PType,
 							 get_state,
 							 [Nidx, jid:tolower(Entity)]) of
-					    {result, State} ->
-						ItemIds = State#pubsub_state.items,
+					    {result, #pubsub_state{items = ItemIds}} ->
 						node_action(Host, PType,
 							    remove_extra_items,
 							    [Nidx, 0, ItemIds]),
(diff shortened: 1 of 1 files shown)


fdie:
The fix solves the following error occuring on remove_user with pubsub subscriptions

<0.3232.409>@mod_pubsub:get_transaction_response/1:4330 Transaction aborted:
        ** exception error: no case clause matching 
                          {pubsub_state,{{<<"3c320c0db3054fe59acc8d533b692b42">>,
                                          <<"xmpp.org">>,<<>>},
                                         492125},
                                        492125,[],member,[]}
           in function  mod_pubsub:node_call/4 (src/mod_pubsub.erl, line 4238)
           in call from mod_pubsub:'-do_transaction/4-fun-0-'/3 (src/mod_pubsub.erl, line 4291)
           in call from ejabberd_sql:execute_bloc/1 (src/ejabberd_sql.erl, line 650)
           in call from ejabberd_sql:run_sql_cmd/4 (src/ejabberd_sql.erl, line 551)
           in call from p1_fsm:handle_msg/10 (src/p1_fsm.erl, line 582)
           in call from proc_lib:init_p_do_apply/3 (proc_lib.erl, line 226)

In mod_pubsub:708 no #pubsub_state record is expected, but a "result" tuple:

					case node_action(Host, PType,
							 get_state,
							 [Nidx, jid:tolower(Entity)]) of
					    {result, State} ->
						ItemIds = State#pubsub_state.items,

PR #3565, opened 2021-04-08
unwrap mucsub and treat payload like a message

Robbilie:
since mucsub messages are wrapped the summary does not detect the body properly. thus checking if its a mucsub message and in that case unwrapping it solves that issue. i assume this is the expected behaviour so this patch brings the behaviour in line with normal push notifications.

   description.

   them as separate PRs.


weiss:
Many thanks for pointing this out and suggesting a fix!  I'll apply a slightly different change, as `mod_push` checks for message bodies in another two or three places.

PR #3566, opened 2021-04-08
allow shared roster group placeholder in mqtt topic

Robbilie:
to allow for a more dynamic access control of topics, even at runtime and since shared roster groups are already part of the acl model i suggest a placeholder for shared roster groups in mqtt topics. the code resolves the respective roster group, checks its existence and the membership of the subscribing/publishing user.


mremond:
Thanks for the PR, this is an interesting feature.

Robbilie:
@mremond i thought so too :P sadly i had to realize that ejabberd only implements a subset of mqtt 5 and the main feature i need, shared subscriptions, isnt supported :(

PR #3577, opened 2021-04-16
Use rebar3 in Travis tests

badlop:
ejabberd can be compiled using rebar2, rebar3 or mix. I have been using rebar3 since some months ago, and it works perfectly for building ejabberd, at least in my experience.

Regarding Travis, I haven't seen any specific performance benefit in Travis using rebar3 instead of rebar2: the testing time just reduces between one to three minutes (15 -> 14 minutes, 18 -> 15 minutes)

But by upgrading Travis from rebar2 to rebar3, we move ejabberd one step forward, and leave rebar2 one step behind. This would make easier, in the future, to drop rebar2 support.

So, here is a proposal to use rebar3 in Travis. Any feedback and opinion is welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,31 +35,32 @@ before_script:
 
 cache:
   directories:
-  - deps/
+  - ~/.cache/rebar3/
 
 script:
   - ./autogen.sh
-  - ./configure --prefix=/tmp/ejabberd --enable-all --disable-odbc --disable-elixir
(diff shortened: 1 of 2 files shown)


badlop:
Now that ejabberd won't use Travis CI, this PR is useless. BTW, the new github actions workflow includes tests for rebar2 and  rebar3.

Issue #3576, opened 2021-04-16
Shared roster group opts name changed to label - breaks when upgraded from 20.04 and older


sarsonj:
## Environment

- ejabberd version: 21.04
## Bug description

We are upgrading after year to new version of Ejabberd (21.04) and we realised, that shared roster group data structure changed (commit has that made this change - e197b25e820ff33390d73551e3e8fac0964d7e2c). To be specific, the attribute "name" in opts changed to "label". This is breaking change for us - we are encoding some information into this field and when there is none, our app don't work correctly.  What worse - the API call `srg_create` still accepts name in its parameter list and also docs refers to `name` field, not `label` (https://docs.ejabberd.im/developer/ejabberd-api/admin-api/). The `label` is even not accepted by `sgr_create`, so that there is no way to create shared roster group using API (with label).

I checked to mod_shared_roster.erl and it looks like that nobody considered migration when made this change. 

So it would be nice at least update docs and `srg_create` API to returns error when `name` is set to avoid confusion and find problem soon.

Now I am thinking how to fix this issue in our production environment. I probably don't want to change ejabberd source files, because I want to upgrade again in the future. But I can change database. I'd like to ask, if it is safe to just update `opts` fields in SQL database - change name to label or even better duplicate name data and have them in both name and label field? Don't it break anything?

badlop:
> the attribute "name" in opts changed to "label".

Right, but that commit maintained support for opts "name", precisely to handle not only new "label", but also old databases without requiring any changes by the system administrator:

https://github.com/processone/ejabberd/commit/e197b25e820ff33390d73551e3e8fac0964d7e2c#diff-0ff381597dbb443bc5fb765d0c78468cebe7a8061f9cb84991276f2de7f77735L1140

Now that you found a problem, I'll check again that code...

> if it is safe to just update opts fields in SQL database - change name to label

Yes. Apparently there's some problem that breaks handling old opts->name, and you can update explicitely your database, even if, in theory, it shouldn't be needed.

> the API call srg_create still accepts name in its parameter list

Oops, I completely forgot about that administration API. I'll check and update them.

sarsonj:
Thanks for clarification, this happen when I create new group with API in new eJabberd and add user to this group with API. Then the roster in XMPP client don't get the old name label. Probably, this case is not handled.

[...]

PR #3578, opened 2021-04-17
Fixed srg_create params and shared roster groups cache issues

sarsonj:
This pull request fixes issues with shared roster groups:

1. fixed srg_create administrative api function - it accepted old name parameter and stores it as `name` into OPTS. The parameter should be now named and stored as `label`
2. it fixes cache issues. The cache was cleaned **before** item was added / removed from shared roster group. So that it can happen (and it happened to us), that user was added to shared roster group, but cache item was reloaded before item was updated in database - so there were old data in cache. I changed the order (first update in DB, then delete cache) and it fixed this inconsistency issue.

This pull request is fixing issue #3576.

### DIFF
--- a/src/mod_admin_extra.erl
+++ b/src/mod_admin_extra.erl
@@ -664,7 +664,8 @@ get_commands_spec() ->
 			"name desc \\\"group1\\\\ngroup2\\\"",
 			module = ?MODULE, function = srg_create,
 			args = [{group, binary}, {host, binary},
-				{name, binary}, {description, binary}, {display, binary}],
+				{label, binary}, {description, binary}, {display, binary}],
+			args_rename = [{name, label}],
 			args_example = [<<"group3">>, <<"myserver.com">>, <<"Group3">>,
 				<<"Third group">>, <<"group1\\\\ngroup2">>],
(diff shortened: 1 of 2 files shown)


Neustradamus:
@sarsonj: Good job!

I hope a quick review of @processone team and without removal ;)

badlop:
Thanks!

I was finally able to get the testing evironment, and I've checked this:
1. installed ejabberd 20.03
2. created mysql & pgsql databases
3. created several shared roster groups
4. upgraded ejabberd to 21.04

Then i confirm:
* A) opts->name are accepted correctly, no need to update the database
* B) srg_create API uses the old "name" argument, which is confusing
* C) newly created groups work correctly, but **name** is only displayed after a server restart

After applying your PR, B) and C) get fixed :)

Neustradamus:
Thanks @badlop for the test and merging.

Thanks a lot to @sarson for your PR :)

badlop:
> Thanks @badlop for the test and merging.
> 
> Thanks a lot to @sarson for your PR :)

That is a useless comment, it adds nothing. It's annoying. You did nothing at all. You just added seperal useless comments. Please remove that comment immediately.

PR #3582, opened 2021-04-19
MySQL Backend Patch for scram-sha512

Nikatik:
scram-sha512 did not work, because serverkey is longer then that array. All passwords was unhashed.

### DIFF
--- a/sql/mysql.new.sql
+++ b/sql/mysql.new.sql
@@ -20,8 +20,8 @@ CREATE TABLE users (
     username varchar(191) NOT NULL,
     server_host varchar(191) NOT NULL,
     password text NOT NULL,
-    serverkey varchar(64) NOT NULL DEFAULT '',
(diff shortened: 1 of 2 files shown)


Neustradamus:
@Nikatik: Good catch! :)

@prefiks: Can you check it?

badlop:
> @Nikatik: Good catch! :)
> 
> @prefiks: Can you check it?

That is a useless comment, it adds nothing. It's annoying. You did nothing at all. You just added several useless comments.

Please, can you remove that comment immediately?

And please, can you NOT add more useless comments?

Thanks!

prefiks:
Yes indeed seems that 64 chars is not enough here, i believe we need at lease 88 chars here [ceil((512/8)/3)*4)].

Neustradamus:
@Nikatik: Can you sign the CLA and the comment of @prefiks?

prefiks:
Thank you

PR #3615, opened 2021-05-28
mod_muc_room: Don't leak owner JIDs

weiss:
Avoid publishing room owner JIDs via the `muc#roominfo` form.


licaon-kter:
Brutal...but fair.

prefiks:
I think this is fine, theoretically you could get that info for public rooms without password other way, so we could expose that info for those, but i think we can live without that information visible.

PR #3652, opened 2021-07-28
PubSub: Bump default value for 'max_items' limit

weiss:
Bump the default value for `mod_pubsub`'s `max_items_node` option, which hard-limits the `max_items` value requested by clients.

These days, use cases such as [microblogging][1] or [XEP-0402][2] may need a large number of items per node.  Bumping the limit makes sure such functionality is properly supported with the default configuration.

[1]: https://xmpp.org/extensions/xep-0277.html
[2]: https://xmpp.org/extensions/xep-0402.html

### DIFF
--- a/include/pubsub.hrl
+++ b/include/pubsub.hrl
@@ -23,7 +23,7 @@
 -define(ERR_EXTENDED(E, C), mod_pubsub:extended_error(E, C)).
 
 %% The actual limit can be configured with mod_pubsub's option max_items_node
--define(MAXITEMS, 10).
+-define(MAXITEMS, 1000).
 
 %% this is currently a hard limit.
 %% Would be nice to have it configurable.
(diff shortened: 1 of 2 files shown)


badlop:
Is this PR still awaiting positive/negative opinions, almost two months later? I guess it needs to be merged and published in order to get more feedback

badlop:
As explained by Holger in the chatroom:

> I submitted them as PRs rather than just pushing the changes because they involve trade-offs and I don't want to make any policy decisions behind the backs of you guys (generally I just push obvious bug fixes or features that—hopefully—don't affect existing functionality).

> having max_items=10 will break real-world use cases now that multi-item PEP seems to become a thing (for Bookmarks and maybe OMEMO and whatnot).

> If anyone is worried about max_items=1000 one could maybe go for some value in between.  My idea behind 1000 was "seems enough for everyone && doesn't really hurt".

mremond:
👍 I agree that this is a change that will help spread new use cases.

PR #3653, opened 2021-07-28
Bump default 's2s_timeout' value

weiss:
Wait for an hour before closing an idle s2s connection.

It's not uncommon for a connection to be idle for longer than ten minutes but less than an hour.  For example, [XEP-0410: MUC Self-Ping][1] suggests a ping interval of fifteen minutes.  A longer idle timeout avoids the latency and log entries associated with constantly re-establishing such connections.

[1]: https://xmpp.org/extensions/xep-0410.html


badlop:
As explained by Holger in the chatroom:

> I submitted them as PRs rather than just pushing the changes because they involve trade-offs and I don't want to make any policy decisions behind the backs of you guys (generally I just push obvious bug fixes or features that—hopefully—don't affect existing functionality).

> `s2s_timeout`, is a trade-off of resource usage (only seems relevant if your server has tens of thousands of s2s peers?) vs. latency.  1h is already the compromise.  Personally I use 24h.

PR #3654, opened 2021-07-29
Fix `i_sr_user_sh_jid_group` index on MySQL new schema

karimhm:
The `i_sr_user_sh_jid_group` index on MySQL new schema uses the wrong table.


badlop:
Right! I was checking recent reports, found [this comment](https://github.com/processone/ejabberd/commit/95fa43aa96514b7e8b77fa7c29d2c0b5b1c1331a#r54157330) and just [applied the fix](aefc374317860aa1e7fd1ed305cbf71b3ce132a8).

Now I see you even provided a PR. I would have saved some time, if I had read this before.

Thanks!

Neustradamus:
@badlop: Your fix link is bad, please test the link and solve it.

badlop:
Link fixed.

PR #3660, opened 2021-08-12
"sort -R" command in captcha-ng.sh is not POSIX - added "shuf" and "cat" as fallback

ballerburg9005:
Some busybox cannot do sort -R (=random). Shuf is an alternative but some systems do not even have that.

This is important for docker containers. 

I am running an 1:1 copy of the official ECS docker image, just build by myself against ARM. This is how I noticed the problem.

Tested and works on all sorts of systems.


badlop:
Great, thanks for reporting and fixing :)

PR #3724, opened 2021-12-07
Remove CTLLOCKDIR (/var/lock/ejabberdctl) from Makefile.in

Flowdalic:
Flock'ing /var/lock/ejabberdctl by ejabberdctl was removed with
f7d4aae64db8 ("Use UUID for ctl node name (#1021)"), however the
according recipies in the Makefile where never removed. This commit
does that.


prefiks:
Thank you, looks fine, so let's merge it.

prefiks:
Ah, i see Holger was faster than me ;)

PR #3727, opened 2021-12-10
Update rebar3 command for `make update`

licaon-kter:
Since https://github.com/erlang/rebar3/commit/8426b2e9d5335265bbad6426ae8399f4a59bb581 rebar3 needs `--all` for `upgrade`

Else it errors out with: `===> Specify a list of dependencies to upgrade, or --all to upgrade them all`

/LE: fix commit link

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -115,7 +115,7 @@ else
 ifeq "$(REBAR_VER)" "3"
   SKIPDEPS=
   LISTDEPS=tree
-  UPDATEDEPS=upgrade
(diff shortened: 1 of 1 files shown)


licaon-kter:
Yes, the pipeline will fail for now until rebar3 will release a new version with this change.

licaon-kter:
rebar3 3.18 is out now...

prefiks:
I commited dc8196693837650c973834ed5f634e1f6e8fa1e8 which adds that flag (or not) depending on version of rebar used.

PR #3757, opened 2022-01-24
Fix typos

kianmeng:


badlop (commented):
Thanks for fixing! It's also good news that those typo fixes don't affect functional code, so they don't require specific upgrade instructions.

kianmeng:
:tada: :tada: :tada: :tada: :tada:

PR #3832, opened 2022-05-31
Support ERL_DIST_PORT option to work without epmd

badlop:
This PR implements an option to disable EPMD and use instead a fixed distribution port. This feature is optional, and disabled by default. The proposed port could be 5210, for example.

It would be great if that feature gets tested, and also the current implementation gets reviewed in case there's a simpler way to achieve the required changes in ejabberdctl.template.

Additionally, this PR includes a commit that enables this feature in the container published in the Github Package registry. It would be great if people with experience in containers mention if that feature is worth enabling, if it makes administration easier, or if it introduces more troubles for administering and clustering than benefits...

Notice:
- Erlang/OTP 23.1 or higher is required to use ERL_DIST_PORT
- `make relive` doesn't support ERL_DIST_PORT, neither rebar3 nor elixir
- To use "ejabberd" script: `ERL_DIST_PORT=5210 _build/dev/rel/ejabberd/bin/ejabberd ping`
- It seems `make install` doesn't work with Elixir's mix since at least ejabberd 21.07, maybe older
- To connect from local "ejabberdctl" to containerized ejabberd:
  export port 5210 from container, and share COOKIE file

Successfully tested with:
- Erlang/OTP 23.1 and 25.0
- rebar 2.6.4
- rebar3 3.14.3 and 3.18
- elixir 1.10.3 and 1.13

Reference:
https://www.erlang.org/blog/otp-23-highlights/
https://blog.erlware.org/epmdlessless/

### DIFF
--- a/.github/container/Dockerfile
+++ b/.github/container/Dockerfile
@@ -61,6 +61,9 @@ RUN export PEM=/opt/ejabberd/conf/server.pem \
         \ncertfiles: \
         \n  - /opt/ejabberd/conf/server.pem' "/opt/ejabberd/conf/ejabberd.yml"
 
+RUN sed -i '/^#ERL_DIST_PORT:/a \ \
+        \nERL_DIST_PORT: 5210' "/opt/ejabberd/conf/ejabberdctl.cfg"
+
 FROM alpine:3.15.4
 ENV HOME=/opt/ejabberd
 ARG VERSION=master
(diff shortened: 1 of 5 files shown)


RomanHargrave:
RE your comments in #3896, I am not currently experiencing any difficulty related to `epmd` in a containerized and clustered configuration. With that having been said (and having read the erlware blog post), I think this could certainly simplify containerized deployments.

badlop:
Ok, in the end I've comitted support for ERL_DIST_PORT, disabled by default in ejabberd itself (30f1e284681db482ef78da6d865ccc79120e4ca3) and also disabled by default in the container (5ee1dc9e8d80d52ec9f3dd3f043dfdd4c90c303a).

PR #3835, opened 2022-06-02
Support @VERSION@ and @SEMVER@ in captcha_cmd option

badlop:
Those macros are useful because the example captcha scripts are copied in a path `ejabberd-$VERSION/priv/bin` tha depends on the ejabberd version number, and changes for each release. Also, depending on the install method (one that uses rebar3, or Elixir's mix), that $VERSION may be in `XX.YY` or in `SEMVER` format (respectively).

Thanks to @weiss for the `semver_to_xxyy/1` function code.


Neustradamus:
@badlop: It will be in 22.10?

badlop:
> @badlop: It will be in 22.10?

Finally, yes

PR #3846, opened 2022-06-12
mod_mix_pam: Provide MIX channels as roster entries via hook

lnjX:
Currently the joined MIX channels of a user are stored in the database of mod_mix_pam, but are not provided to the user in any way. This change subscribes to the hook of mod_roster and provides the joined channels.

TODO:
- [x] Add roster pushes
- [x] Only provide MIX subelements when `<mix-annotate/>` is set
  - [x] in roster get
  - [x] in roster pushes: Requires to store annotate setting for the session (Need to find out how)
- [x] Change roster hook to use `#roster_item{}` instead of `#roster{}` so the roster record can stay how it is (without mix participant ID). This way no empty field for the mix participant id is saved in the roster mnesia db.

### DIFF
--- a/include/mod_roster.hrl
+++ b/include/mod_roster.hrl
@@ -28,7 +28,8 @@
     ask = none                             :: ask() | '_',
     groups = []                            :: [binary()] | '_',
     askmessage = <<"">>                    :: binary() | '_',
-    xs = []                                :: [fxml:xmlel()] | '_'
+    xs = []                                :: [fxml:xmlel()] | '_',
+    mix_participant_id = <<>>              :: binary() | '_'
 }).
 
 -record(roster_version,
--- a/include/mod_roster.hrl
+++ b/include/mod_roster.hrl
@@ -28,8 +28,7 @@
     ask = none                             :: ask() | '_',
     groups = []                            :: [binary()] | '_',
     askmessage = <<"">>                    :: binary() | '_',
-    xs = []                                :: [fxml:xmlel()] | '_',
-    mix_participant_id = <<>>              :: binary() | '_'
+    xs = []                                :: [fxml:xmlel()] | '_'
 }).
 
 -record(roster_version,
--- a/src/ejabberd_c2s.erl
+++ b/src/ejabberd_c2s.erl
@@ -777,9 +777,9 @@ broadcast_presence_unavailable(#{jid := JID, pres_a := PresA} = State, Pres,
 		    Roster = ejabberd_hooks:run_fold(roster_get, LServer,
 						     [], [{LUser, LServer}]),
 		    lists:foldl(
-			fun(#roster{jid = LJID, subscription = Sub}, Acc)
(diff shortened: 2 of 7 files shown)


lnjX:
@badlop Do you've got an idea for how to store the `<mix-annotate/>` setting per session? Is that easily possible without using the database?

lnjX:
:partying_face:

badlop:
Umm, there's a problem in commit 76aa98f21e33a106c461671366af5119479e7182, tt breaks mod_shared_roster. Can you take a look?

It's easy to reproduce, for example:
1. Empty database, default configuration, start ejabberd
2. Register an account and login (admin@localhost)
3. That user adds a contact to his roster (user3@localhost)
4. mod_shared_roster immediately crashes:
```
2022-08-12 12:55:09.951716+02:00 [error] Hook roster_get crashed when running mod_shared_roster:get_user_roster/2:
** exception error: {badrecord,
                      {roster_item,
                          {jid,<<"user3">>,<<"localhost">>,<<>>,<<"user3">>,
                              <<"localhost">>,<<>>},
                          <<>>,[],none,subscribe,undefined}}
   in function  mod_shared_roster:'-get_user_roster/2-fun-2-'/2 (/home/bernar/e/git/ejabberd/src/mod_shared_roster.erl, line 206)
   in call from lists:mapfoldl_1/3 (lists.erl, line 1462)
   in call from mod_shared_roster:get_user_roster/2 (/home/bernar/e/git/ejabberd/src/mod_shared_roster.erl, line 204)
   in call from ejabberd_hooks:safe_apply/4 (/home/bernar/e/git/ejabberd/src/ejabberd_hooks.erl, line 320)
   in call from ejabberd_hooks:run_fold1/4 (/home/bernar/e/git/ejabberd/src/ejabberd_hooks.erl, line 301)
   in call from mod_roster:roster_version/2 (/home/bernar/e/git/ejabberd/src/mod_roster.erl, line 236)
   in call from mod_roster:push_item/3 (/home/bernar/e/git/ejabberd/src/mod_roster.erl, line 499)
   in call from mod_roster:process_subscription/6 (/home/bernar/e/git/ejabberd/src/mod_roster.erl, line 647)
** Arg 1 = [{roster_item,{jid,<<"user3">>,<<"localhost">>,<<>>,<<"user3">>,
                              <<"localhost">>,<<>>},
                         <<>>,[],none,subscribe,undefined}]
** Arg 2 = {<<"admin">>,<<"localhost">>}
```

prefiks:
badlop: I think that mod_shared_roster was not recompiled after xmpp changes (i think this record is correct by current definition).

lnjX:
@badlop I think I've found the issue, https://github.com/processone/ejabberd/pull/3882 should fix it!

badlop:
`make distclean` and removing rebar.lock didn't help, as the problem is in that hook that wasn't updated to the new event arguments. #3882 solves it, thanks!

PR #3853, opened 2022-06-26
mod_mix_pam: Display joined channels on webadmin page

lnjX:
Makes it possible to see which channels a user has joined (inspired by the normal roster entries). Useful for debugging.

image
image

### DIFF
--- a/src/mod_mix_pam.erl
+++ b/src/mod_mix_pam.erl
@@ -31,11 +31,15 @@
 -export([bounce_sm_packet/1,
 	 disco_sm_features/5,
 	 remove_user/2,
-	 process_iq/1]).
+	 process_iq/1,
+	 webadmin_user/4,
+	 webadmin_page/3]).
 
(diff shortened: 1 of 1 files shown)


badlop:
Looks great, thanks!

Now that I looked at MIX, I've updated the MIX tutorial in the documentation site https://docs.ejabberd.im/tutorials/mix-010/

BTW, in case you missed this: in a small testing server running a recent ejabberd, there's a way to look at raw Mnesia content in webadmin: in http://localhost:5280/admin/node/ejabberd@localhost/db/ the `Elemens` are links to raw content.

lnjX:
> BTW, in case you missed this: in a small testing server running a recent ejabberd, there's a way to look at raw Mnesia content in webadmin: in http://localhost:5280/admin/node/ejabberd@localhost/db/ the Elemens are links to raw content.

Ah I actually wasn't aware of it, thanks :)

And thanks for updating the docs :)

PR #3854, opened 2022-06-26
Handle newer versions of the mix namespaces

JBBgameich:
As far as I can see, there are no other changes to be made to add compatibility with the newer namespaces:
`mix:pam:1` was bumped to make archiving optional, which afaik ejabberd did not implement anyway. If it did before, it would still be compliant
`mix:pam:2` was bumped, in order to communicate that mix:core was bumped
`mix:core:1` itself was bumped to adapt the text to reflect that archiving was made optional.


lnjX:
PR itself is ready, but just requires https://github.com/processone/xmpp/pull/65 to be merged (and tagged). :)

badlop:
> PR itself is ready, but just requires processone/xmpp#65 to be merged (and tagged). :)

Usually those libraries are tagged only a few days before next ejabberd release...

You don't need to wait for it to be tagged: you can update rebar.config and mix.exs to use a recent commit.

Example usage: https://github.com/processone/ejabberd/commit/8f8de0403bab0a819420db69bb44bdb2a474d137

In your case, the ref would be `"5b22f0c0ae189d0eda8ba469c0629041c1777857"`, which is the most recent `xmpp` commit.

lnjX:
> The good news: you don't need to wait for it to be tagged: you can update rebar.config and mix.exs to use a recent commit.

That's good to know! Changed it, so PR should be ready now. For mix.exs I had to set the specific git repo with override true, hope that's fine. :)

PR #3860, opened 2022-07-08
Postgres schema updates

mzealey:
Please see the individual commit messages. I have not attempted to apply this to the 'new' schema and there are a couple of TODOs where I think something should be NOT NULL but I'm not 100% sure.

Combined with some server-side postgres performance optimizations this schema is supporting a cluster containing millions of simultaneous active users on relatively small hardware.

### DIFF
--- a/sql/pg.sql
+++ b/sql/pg.sql
@@ -47,11 +47,10 @@ CREATE TABLE rosterusers (
     server character(1) NOT NULL,
     subscribe text NOT NULL,
     "type" text,
-    created_at TIMESTAMP NOT NULL DEFAULT now()
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    PRIMARY KEY (username, jid)
 );
 
-CREATE UNIQUE INDEX i_rosteru_user_jid ON rosterusers USING btree (username, jid);
-CREATE INDEX i_rosteru_username ON rosterusers USING btree (username);
 CREATE INDEX i_rosteru_jid ON rosterusers USING btree (jid);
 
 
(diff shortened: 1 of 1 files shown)


licaon-kter:
@mzealey 
> some server-side postgres performance optimizations

Do tell more. :wink:

mzealey:
> @mzealey
> 
> > some server-side postgres performance optimizations
> 
> Do tell more. wink

The key ones are `commit_delay=10000` and `synchronous_commit=off` which accept a little bit of potential data loss for a massive reduction in latency/iops.

> LE: also, why remove so many indexes?

Sorry only saw this now - see commit message for https://github.com/processone/ejabberd/pull/3860/commits/22d49ad2fe6bd6e7cc3217325404c74c31178e2d about the rationale behind this; let me know if you need more detail?

nosnilmot:
The most obviously beneficial part of this PR is the elimination of redundant indexes. Those changes are applicable to ALL schemas - old and new, and for all DB types. Commit https://github.com/processone/ejabberd/pull/3980/commits/06ffe995e1db6168c15200028360517c54212ed4 in PR #3980 addressed those.

It could be argued either way which of UNIQUE INDEX and PRIMARY KEY constraints is more explicit/logical/understandable, but again any change should be made across all schemas for consistency. I'd argue that there are limited benefits and sticking with the status-quo is preferred here. Any change would also have knock-on effects on the schema upgrade scripts and ongoing maintenance for existing installations.

Changing columns from SERIAL to BIGSERIAL is another reasonable suggestion (and again needs to apply to both old and 'new' schemas). Commit https://github.com/processone/ejabberd/pull/3980/commits/4f0e426a12aadc6a7249e6a8d6fe62cdb0579cd7 in PR #3980 addressed this.

I'm not convinced the fillfactor performance changes are useful to include in a general purpose schema - premature optimization can be less than helpful, and I feel these might be workload dependent.

mzealey:
Looks reasonable thanks. I don't think the `fillfactor` is particularly bad - Pg sets it (I believe to 90%) on btree indexes anyway, so setting it at 90% on a table which has high volume of UPDATEs will just allow much less disk IO (via HOT mechanism) at a slight cost of disk space. I don't see how it's ever going to be detrimental.

Neustradamus:
To follow

Neustradamus:
@mzealey: Can you look for pg.new.sql too?

There is a second PR too:
- https://github.com/processone/ejabberd/pull/4113

PR #3863, opened 2022-07-18
move spool dir to make it fully configurable per ejabberdctl.cfg

duritong:
In the packaged rpm the spool dir is set to:
```
: "${SPOOL_DIR:="/opt/ejabberd/database/$ERLANG_NODE"}"
```

However, `$ERLANG_NODE` is effectively set later (now in line 67), which effectively makes spool dir always in `...../ejabberd@localhost`

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -59,7 +59,6 @@ done
 # define ejabberd variables if not already defined from the command line
 : "${CONFIG_DIR:="{{config_dir}}"}"
 : "${LOGS_DIR:="{{logs_dir}}"}"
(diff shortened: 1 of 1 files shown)


weiss:
Good catch, thanks.

Yes this should be fixed, I'm just a little unsure whether it might make sense to move `CONFIG_DIR` and `LOGS_DIR` down as well, for consistency/cosmetics.  The only behavior change I'd see is that you cannot do `LOGS_DIR="$LOGS_DIR/subdir"` in `ejabberdctl.cfg` then, which seems uninteresting.  But I'd need another coffee to be sure.

PR #3865, opened 2022-07-19
Add log_burst_limit_* options

mzealey:
On our ejabberd deployment we were sometimes seeing more than 500
msgs/sec of legitimate traffic, however this was getting silently
dropped. Provide config options to enable this limit to be configured
from the config file.


prefiks:
Looks good, do you have something else to add here or is it ready to merge?

mzealey:
I believe this should be complete. Annoying about `ejabberdctl` hackiness but that can perhaps be addressed in a different patch by someone who understands erlang better than I.

PR #3882, opened 2022-08-12
mod_shared_roster: Fix wrong hook type #roster{} (now #roster_item{})

lnjX:
The hook type was changed from #roster{} to #roster_item{} by me in the
recent MIX changes. Unfortunately I've overseen this one (and unit tests + dialyzer passed).


badlop:
This solves the problem. Thanks!

PR #3913, opened 2022-10-03
fix table creation on a running cluster

fdie:
On a 2 nodes A & B running cluster, the "ejabberd_mnesia:create" on node A will create the mnesia table, but the same "ejabberd_mnesia:create" call on node B will trigger a table update since the table is "known".

This second call fails on the following call :

ejabberd_mnesia:mnesia_op(change_table_copy_type, [mytable,'ejabberd-cluster@my_xmpp.com',ram_copies]) => {aborted, {badarg,mytable,unknown}}

The function "ejabberd_mnesia:change_table_copy_type" manages configuration changes but requires an existing local copy, if the storage type is "unknown", create this copy with "mnesia_op(add_table_copy, [Name, node(), NewType])" instead fixes the issue.


prefiks:
Looks good, let's merge

PR #3945, opened 2022-11-16
make-binaries: Bump zlib to 1.2.13

nosnilmot:
Update zlib to 1.2.13 to address CVE-2022-37434

Also change zlib download to use archive URL so builds do not fail when specified version is no longer the current version, and therefore no longer available to download at default location

And reduce log output:
 - quieter curl downloads
 - quiet build of crosstool-ng
 - disable crosstool-ng progress bar

### DIFF
--- a/tools/make-binaries
+++ b/tools/make-binaries
@@ -260,6 +260,7 @@ create_common_config()
 		CT_LINUX_V_3_16=y
 		CT_GLIBC_V_2_17=y
 		CT_GLIBC_KERNEL_VERSION_NONE=y
+		CT_LOG_PROGRESS_BAR=n
 	EOF
 }
 #.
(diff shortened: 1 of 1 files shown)


weiss:
Thanks.

You're editing the crosstool-NG configuration to update their zlib version as well.  Is that necessary to make the build succeed?  Otherwise I'd suggest omitting that change (the toolchain's zlib isn't actually used by the resulting binaries).

nosnilmot:
> You're editing the crosstool-NG configuration to update their zlib version as well. Is that necessary to make the build succeed? Otherwise I'd suggest omitting that change (the toolchain's zlib isn't actually used by the resulting binaries).

yes, it is necessary, otherwise crosstool-ng also fails to download zlib-1.2.12 because it's also looking for it at https://zlib.net/zlib-1.2.12.tar.xz (https://github.com/crosstool-ng/crosstool-ng/issues/1832).
I could omit the CT_ZLIB_VERSION change and just update CT_ZLIB_MIRRORS to get 1.2.12 from the archives (fossils) but if making changes in that area we might as well use the version that does not have a vulnerability :)

weiss:
Ok, thanks 👍🏻

PR #3970, opened 2023-01-08
Fix portability issue with some shells

triaxx:
Some shells do not implement a test command which supports the '==' operator. For the sake of compatibility with the greatest number of environments, it is preferable to use the '=' operator.


prefiks:
Thank you

Issue #3695, opened 2021-10-10
vcard_iq_set crashed when running mod_vcard:vcard_iq_set/1 


licaon-kter:
## Environment

- ejabberd version: d205e6ff1ff32c075afea8b3189edc5f3c0c1ab0 or later, I guess
- Erlang version: `Erlang (SMP,ASYNC_THREADS) (BEAM) emulator version 11.1.8`
- OS: Debian Bullseye arm64
- Installed from: source

## Configuration (only if needed): grep -Ev '^$|^\s*#' ejabberd.yml

https://gist.github.com/licaon-kter/df2af81dbd341837999a38502de79b89

## Errors from error.log/crash.log

https://gist.github.com/licaon-kter/fea9a0ace5195b5c2f9cf965a77230a7

licaon-kter:
Postgres log says:
```
ejabberd@ejabberd STATEMENT:  INSERT INTO vcard_search(username, lusername, server_host, fn, lfn, "family", lfamily, given, lgiven, middle, lmiddle, nickname, lnickname, bday, lbday, ctry, lctry, locality, llocality, email, lemail, orgname, lorgname, orgunit, lorgunit) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25) ON CONFLICT (lusername, server_host) DO UPDATE SET username=$26, fn=$27, lfn=$28, "family"=$29, lfamily=$30, given=$31, lgiven=$32, middle=$33, lmiddle=$34, nickname=$35, lnickname=$36, bday=$37, lbday=$38, ctry=$39, lctry=$40, locality=$41, llocality=$42, email=$43, lemail=$44, orgname=$45, lorgname=$46, orgunit=$47, lorgunit=$48
```

PR #3982, opened 2023-01-21
SQL schema migration fixes and testing

nosnilmot:
* Un-deprecate ejabberd_config:set_option/2
  This might be controversial, but it was necessary for the next item. Is there another way to change config settings at runtime?
* New schema migration 'update_sql' improvements
  -   Check that server_host column does not already exist before addding it and making other changes to table (update_sql becomes idempotent, yay!)
  -   Check that indexes exist before dropping them (some are historical and are not created in more recent deployments), elminating spurious errors from logs
  -   Update new_sql_schema config after migration, to allow minimizing downtime during migrations (and help with automated testing)
* Add ability to run tests on upgraded DB
  And enable them in github actions - only run if changes are detected in schema or update_sql code (it would be good to also run this at least once pre-release, but it might be overkill to run for every change)
* Add support for running tests on MS SQL
  And enable them in github actions
* Fix a long standing bug in new schema on PostgreSQL
  Same bug as #3695 which was fixed for new installations but not for upgrades. The fix for any existing impacted installations is the same:

```sql
ALTER TABLE vcard_search DROP CONSTRAINT vcard_search_pkey;
ALTER TABLE vcard_search ADD PRIMARY KEY (server_host, lusername);
```

finding that bug _almost_ made all the effort to enable automated schema upgrade testing worthwhile

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,8 +58,19 @@ jobs:
         wget https://github.com/processone/ejabberd/raw/21.12/rebar3
         chmod +x rebar3
 
+    - name: Install MS SQL Server
+      run: |
+        docker run -d -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=ejabberd_Test1" \
+                   -v $(pwd)/test/docker/db/mssql/initdb/initdb_mssql.sql:/initdb_mssql.sql:ro \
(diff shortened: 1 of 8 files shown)


Neustradamus:
To follow :)

prefiks:
Took me some time to review it, but it all looks sensible. Let's merge it.

Issue #3330, opened 2020-07-13
MUC/Sub Persist Roles of a Subscriber


partikmadan:
**Is your feature request related to a problem? Please describe.**
Currently roles are not part of subscribers list and not persisted in db.

**Describe the solution you'd like**
To make the role persistent just like affiliation.

**Describe alternatives you've considered**
Will look into updating the source code if there is no other way.

**Additional context**
`As to implement a modern chat app i need to distinguish visitor from participants in a muc-sub enabled muc room. Currently i am not able to think of any solution for that.`

badlop:
> Currently roles are not part of subscribers list and not persisted in db.

Right, that's exactly the definition of "role", it's like affiliation but non-persistent:

> Role
> A temporary position or privilege level within a room, distinct from a user's long-lived affiliation with the room; the possible roles are "moderator", "participant", and "visitor" (it is also possible to have no defined role). A role lasts only for the duration of an occupant's visit to a room
> https://xmpp.org/extensions/xep-0045.html#terms-general

Then you propose:

> To make the role persistent just like affiliation.

If I understand this correctly, this would go against the XEP, right? That is perfectly acceptable if you control both the server and the clients... But I doubt that would be acceptable for an open XMPP server.

alepeino:
hi @badlop . According to the XEP
> Roles are temporary in that they do not necessarily persist across a user's visits to the room and MAY change during the course of an occupant's visit to the room. An implementation MAY persist roles across visits and SHOULD do so for moderated rooms (since the distinction between visitor and participant is critical to the functioning of a moderated room).

Do you know if there are plans to implement the persistence of roles in Ejabberd? Currently if a moderator revokes voice from a user and the user leaves and re-joins the room, the role is back at "participant". Do you know of any other way to implement this as to honor the moderator's decission?

badlop:
Just for curiosity, I've implemented it in https://github.com/badlop/ejabberd/commit/58ca0a40531a9aab8337d28019c34dd600375ece so you can apply that and try it

alepeino:
thanks! I will give it a look

badlop:
This feature goes against the XEP, and only one person asked for it, so it will not be included in ejabberd. Anybody interested, can find the patch in https://github.com/badlop/ejabberd/commit/58ca0a40531a9aab8337d28019c34dd600375ece

alepeino:
[...]

PR #4010, opened 2023-03-19
Persist `none` role for outcasts

based-a-tron:
`none` roles should be persisted for banned users. I totally forgot about this, my bad.


badlop:
That PR follows 4d3875f4f and a8121cd7e related to issue #3330, and were included in ejabberd 22.10.

PR #4071, opened 2023-08-03
Dockerfile: Include an entrypoint script for ejabberdctl

sando38:
Hi @badlop 

The entrypoint script serves as a placeholder, so that admins can replace it with an actual entrypoint script, e.g. for initiating ejabberd clusters, through mounting it to the path `/usr/local/bin/run.sh`.

This PR does not change anything of the existing functions of the image.

Motivation:
I would like to use the image as basis for the [helm-chart](https://github.com/sando38/helm-ejabberd) with these [scripts](https://github.com/sando38/helm-ejabberd/tree/main/image/78f81de252dc932cd47b91d1a84ca8e8f0647498/scripts) and not rely on a custom image in the end.

### DIFF
--- a/.github/container/Dockerfile
+++ b/.github/container/Dockerfile
@@ -126,6 +126,10 @@ RUN home_root_dir=$(echo $HOME | sed 's|\(.*\)/.*|\1 |') \
         \nexport SPOOL_DIR=/$HOME/database \
         \nexec /$(find $home_root_dir -name ejabberdctl) \"\$@\"" \
             > usr/local/bin/ejabberdctl \
+    && echo -e \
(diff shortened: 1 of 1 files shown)


sando38:
Deleted the branch by accident, will open again if necessary.

Neustradamus:
@badlop, @prefiks, @mremond: @sando38 needs to create a new PR?

sando38:
I think, I don't. For my particular case, I discovered two options to override `entrypoint` and `cmd`:
https://www.ithands-on.com/2021/11/kubernetes-101-overriding-entrypoint.html

PR #4088, opened 2023-09-21
Improve get_roster command result: show groups as a list

badlop:
This PR improves the `get_roster` command, and fixes ejabberdctl to fully support that command and other ones like that.

This changes the API, so the release notes should describe it. For example:

---

### Changes in `get_roster` command

There are some changes in the result output of the `get_roster` command defined in `mod_admin_extra`:
- `ask` is renamed to `pending`
- `group` is renamed to `groups`
- the new `groups` is a list with all the group names
- a contact that is in several groups is now listed only once, and the groups are properly listed.

For example, let's say that `admin@localhost` has two contacts: a contact is present in two groups (`group1` and `group2`), the other contact is only present in a group (`group3`).

The [old get_roster command in ejabberd 23.04](https://docs.ejabberd.im/archive/23_04/admin-api/#get-roster) and previous versions was like:

```
$ ejabberdctl get_roster admin localhost
jan@localhost jan   none    subscribe       group1
jan@localhost jan   none    subscribe       group2
tom@localhost tom   none    subscribe       group3
```

The new [get_roster command](https://docs.ejabberd.im/developer/ejabberd-api/admin-api/#get-roster) in ejabberd 23.XX and newer versions returns as result:

```
$ ejabberdctl get_roster admin localhost
jan@localhost jan   none    subscribe       group1;group2
tom@localhost tom   none    subscribe       group3
```

Notice that the `ejabberdctl` command-line tool since now will represent list elements in results separated with `;`

### DIFF
--- a/src/ejabberd_ctl.erl
+++ b/src/ejabberd_ctl.erl
@@ -335,7 +335,7 @@ call_command([CmdString | Args], Auth, _AccessCommands, Version) ->
 							ArgsFormatted,
 							CI2,
 							Version),
-	    format_result(Result, ResultFormat);
+	    format_result_preliminary(Result, ResultFormat);
 	{'EXIT', {function_clause,[{lists,zip,[A1,A2|_], _} | _]}} ->
 	    {NumCompa, TextCompa} =
 		case {length(A1), length(A2)} of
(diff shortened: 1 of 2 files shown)


Neustradamus:
Excellent!

PR #4113, opened 2023-11-03
Convert indexes to primary keys

mzealey:
We are starting to use `pglogical` for cross-site replication, and it requires all tables to have a primary key. The following patch shows how this can be done using the most basic pg schema - I know there are various other pg schema iterations elsewhere in the code-base. In postgres a pk is identical to a unique key where all columns are `NOT NULL`, but declaratively more meaningful.

As most of these PK's are replacing existing non-unique btree indexes there is minimal disk space/iops impact. Having a PK constraint in place likely makes the tables more efficient by enforcing a `NOT NULL` constraint on a number of columns which don't have this set at present (but should do, for example some of the sequence-based columns).

```
alter table archive add primary key (id); 
alter table privacy_list_data add primary key (id); 
alter table caps_features add primary key (node, subnode, feature); 
alter table pubsub_node_option add primary key (nodeid, name);           
alter table pubsub_node_owner add primary key (nodeid); 
alter table rostergroups add primary key (username, jid); 
alter table spool add primary key (seq); 
 
drop index if exists pk_rosterg_user_jid; 
drop index if exists i_privacy_list_data_id; 
drop index if exists i_pubsub_node_option_nodeid; 
drop index if exists i_pubsub_node_owner_nodeid; 
drop index if exists i_caps_features_node_subnode; 
```


licaon-kter:
And `pg.new.sql` too?

Neustradamus:
@mzealey: Can you look for pg.new.sql too?

There is a second PR too:
- https://github.com/processone/ejabberd/pull/3860

Neustradamus:
@mzealey: Any progress on it?

PR #4256, opened 2024-07-12
Fix template error for elixir-enabled applications.

slezakattack:
Hello, I have an Elixir-based app that is pulling ejabberd as a dependency and noticed when I run the app as a release, the console logs are missing the messages:

```
 2024-07-12 14:17:24.108891-06:00 [info]
 2024-07-12 14:17:24.142699-06:00 [info]
 2024-07-12 14:17:24.170040-06:00 [info]
 2024-07-12 14:17:24.172795-06:00 [info]
 2024-07-12 14:17:24.225542-06:00 [info]
 2024-07-12 14:17:24.225665-06:00 [warning]
```
I realized that there is a bug in the template that uses `message` when it needs to be `msg` (as seen in the Erlang documentation https://www.erlang.org/doc/apps/kernel/logger_formatter#t:template/0)

After making this change, I get the following, correct logs:
```
 2024-07-12 14:45:59.809550-06:00 [info] Loading configuration from ejabberd.yml
 2024-07-12 14:45:59.842689-06:00 [info] Configuration loaded successfully
 2024-07-12 14:45:59.867775-06:00 [info] Got no NOTIFY_SOCKET, notifications disabled
 2024-07-12 14:45:59.870429-06:00 [info] Building language translation cache
 2024-07-12 14:45:59.920871-06:00 [info] Loading modules for localhost
 2024-07-12 14:45:59.920980-06:00 [warning] Mnesia backend for mod_mam is not recommended: it's limited to 2GB and often gets corrupted when reaching this limit. SQL backend is recommended. Namely, for small servers SQLite is a preferred choice because it's very easy to configure.
```

### DIFF
--- a/src/ejabberd_logger.erl
+++ b/src/ejabberd_logger.erl
@@ -370,7 +370,7 @@ console_template() ->
         andalso
         'Elixir.System':version() >= <<"1.15">> of
         true ->
-            [date, " ", time, " [", level, "] ", message, "\n"];
+            [date, " ", time, " [", level, "] ", msg, "\n"];
         false ->
             [time, " [", level, "] " | msg()]
     end.


badlop:
Thanks for fixing it!

badlop:
Bad news: after applying your patch, ejabberd fails to log messages when used stand-alone. Apparently, ejabberd uses [Elixir's Logger.Formatter](https://hexdocs.pm/logger/1.17.2/Logger.Formatter.html), which [uses `message`](https://hexdocs.pm/logger/1.17.2/Logger.Formatter.html#module-formatting) as originally implemented, not `msg` as used by Erlang/OTP's logger.

I've done a very sple test, and your patch is not needed at all:

1. Create an app
```sh
mix new myapp
cd myapp
vim mix.exs
```
2. Set ejabberd as a dependency:
```elixir
  defp deps do
    [
      {:ejabberd, "~> 24.6"},
    ]
  end
```
3. Get dependencies, compile, and setup ejabberd configuration file:
```sh
mix deps.get
mix compile
wget https://raw.githubusercontent.com/processone/ejabberd/master/ejabberd.yml.example
mv ejabberd.yml.example ejabberd.yml
```
4. Finally start a console, ejabberd 24.6 starts and logs correctly
```sh
mix
2024-07-15 12:26:35.493 [info] Loading configuration from ejabberd.yml
2024-07-15 12:26:35.985 [info] Creating Mnesia ram table 'route'
...
2024-07-15 12:26:36.400 [info] ejabberd 24.6.0 is started in the node :nonode@nohost in 0.93s
2024-07-15 12:26:36.429 [info] Start accepting TLS connections at [::]:5223 for :ejabberd_c2s
...
```

I'll have to revert this PR as it is now.

> I have an Elixir-based app that is pulling ejabberd as a dependency

How did you integrate ejabberd into your elixir application? Maybe your application, or some dependency, is changing the logger configuration from Elixir's one to OTP's one?

Can you provide a minimal example that allows to reproduce your problem?

It would be great if your specific scenario can be detected when ejabberd is starting in console_template, and set the proper log template strings for your specific case.

slezakattack:
> Can you provide a minimal example that allows to reproduce your problem?

Yes, if you take the example you currently have and instead do
```bash
mix release
./_build/dev/rel/myapp/bin/myapp start
```
This would reproduce the issue. It only seems to occur during a release which makes me wonder if it has something to do with application startup order (i.e. erlang logger starts before Elixir.Logger?). I'm new to Elixir so I'm not entirely sure how it comes together along with the Erlang libraries..

badlop:
Aha! Now I could reproduce the problem.

I've added a small fix that works great for me, both in standard ejabberd start methods, and also works using the method you described.

Can you try it and confirm it's fixed? Either apply the change yourself, or download the raw file and replace it.
https://raw.githubusercontent.com/badlop/ejabberd/dependabot/src/ejabberd_logger.erl

slezakattack:
Awesome, yes just tested this out and it seems to work. Thank you!

PR #4259, opened 2024-07-18
[fix] fix mysql.sql archive origin_id

heyanyanchina123:


prefiks:
Thank you

weiss:
```sql
ALTER TABLE archive ADD COLUMN origin_id origin_id(191) NOT NULL DEFAULT '';
```

Shouldn't the type be `varchar(191)` instead?

prefiks:
Ah, indeed, i will change that.

prefiks:
Changed in d2a3fe3ed2956d392fdba9d0d9e713c1bd8169ff

Issue #4178, opened 2024-03-19
[BUG] MS SQL database backend reason: {port_exit,memory_allocation_failed} with Erlang >25.3


sando38:
## Environment

- ejabberd version: 24.02
- Erlang version: `26.2`
- OS: Linux (Debian12) and Alpine Linux 3.19
- Installed from: source | ghcr.io/processone/ejabberd:24.02

## Configuration (only if needed): grep -Ev '^$|^\s*#' ejabberd.yml

```yaml
loglevel: info
```

## Errors from error.log/crash.log

<details>

```
2024-03-16T20:39:12.6496590Z 2024-03-16 20:32:36.414879+00:00 [info] <0.423.0>@gen_mod:start_modules/0:141 Loading modules for 8310292872-12.example.com
2024-03-16T20:39:12.6498391Z 2024-03-16 20:32:36.449292+00:00 [info] <0.374.0>@ejabberd_mnesia:create/2:270 Creating Mnesia ram table 'pubsub_last_item'
2024-03-16T20:39:12.6500208Z 2024-03-16 20:32:36.461769+00:00 [info] <0.374.0>@ejabberd_mnesia:create/2:270 Creating Mnesia ram table 'mod_register_ip'
2024-03-16T20:39:12.6502116Z 2024-03-16 20:32:36.488646+00:00 [info] <0.608.0>@mod_stun_disco:parse_listener/1:626 Going to offer STUN service: 10.244.3.7:3478 (udp)
2024-03-16T20:39:12.6504254Z 2024-03-16 20:32:36.551778+00:00 [info] <0.374.0>@ejabberd_mnesia:create/2:270 Creating Mnesia ram table 'muc_online_room'
2024-03-16T20:39:12.6506198Z 2024-03-16 20:32:36.643316+00:00 [error] <0.649.0> ODBC: received unexpected info: {tcp_closed,#Port<0.91>}
2024-03-16T20:39:12.6507055Z 
2024-03-16T20:39:12.6508629Z 2024-03-16 20:32:36.643508+00:00 [error] <0.632.0>@ejabberd_sql:check_error/2:1404 SQL query 'select name, opts from muc_room where host='conference.8310292872-12.example.com'' failed: connection_closed
[...]

PR #4261, opened 2024-07-20
[Dockerfile] Switch to wolfi/glibc based image

sando38:
Hello ejabberd team, 

@badlop as discussed some time ago, during a simple `processone/rtb` test with the Alpine/musl-libc based container image, I ran into issues when spawning 5-10k users on the container. After digging arround and spending some time, what effectively solved all issues was just using a glibc based variant and the container became well performing.

Switching to the proposed [wolfi/os](https://github.com/wolfi-dev/os) image has the following qualities:
* glibc based container image 
* Alpine-like system, up-to-date package, but based on glibc
* still small image sizes compared to Debian-slim and Co.

Additionally, the github workflows could drop the binary builds for `musl-libc` which have been the work-around also for the QEMU bug blocking building `arm64` variants.

Therefore, this PR includes:
* Switched Dockerfile to use `wolfi` instead of `Alpine`
* Updated the github workflow files to use the output of the installer workflow to build container `package` images

I was not yet able to test the workflows, maybe need to do some fine tuning here when they have been triggered.

Thanks and until soon!
Saarko

### DIFF
--- a/.github/container/Dockerfile
+++ b/.github/container/Dockerfile
@@ -1,7 +1,7 @@
 #' Define default build variables
 ## source ARGs
 ARG OTP_VSN='27'
-ARG ELIXIR_VSN='1.16.3'
+ARG ELIXIR_VSN='1.17.2'
 ## general ARGs
 ARG UID='9000'
 ARG USER='ejabberd'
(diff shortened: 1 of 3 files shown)


badlop:
> I was not yet able to test the workflows, maybe need to do some fine tuning here when they have been triggered.

Great! I've added commits that solve workflow run, tarball download, and use same stable Erlang version than the nbinary installers. See this branch: https://github.com/badlop/ejabberd/commits/wolfi/

Now the workflow runs, the container image is generated, and looks great at least with some basic manual testing

```
$ podman run --name ejabberd -it -p 5222:5222 -p 5280:5280 ghcr.io/badlop/ejabberd:wolfi live
...
14:17:03.953 [info] ejabberd 24.7.13 is started in the node :ejabberd@localhost in 1.26s

$ podman exec -it ejabberd ejabberdctl register admin localhost asd
User admin@localhost successfully registered
```

And a client can login

sando38:
Cool, the only thing [which will not work](https://github.com/badlop/ejabberd/commit/eaaf41ca4d2dd3ae9f4416b1a86ce1ee8ba11789) is the specification of the `OTP_VSN`, because Wolfi OS packages only contain the major version in their package name:
https://github.com/wolfi-dev/os/blob/main/erlang-27.yaml

So you would need to set `OTP_VSN='26'` to keep the same stable major version when using `METHOD=direct` :) Probably I need to create another PR in the Wolfi repository to include the ODBC package for the `erlang-26-odbc` package. We need to keep in mind, however, that with the `METHOD=direct` the MS SQL will not work (#4178)

Neustradamus:
To follow this PR.

PR #4265, opened 2024-08-15
Mix release fails with logger application error

slezakattack:
When attempting a `mix release` on a new mix project, I get the following error:
```bash
** (Mix) :logger is listed both as a regular application and as an included application
```
Repro steps:
1. `mix new ejabberd_mix_release_issue`
2. Within `mix.exs`, add `{:ejabberd, "~> 24.7"}` in the list of deps.
3. `mix deps.get`
4. `mix compile`
5. `mix release`

Updating the `mix.exs` within ejabberd seems to fix the issue as I'm able to generate a release and the application is able to start up (provided there's an `ejabberd.cfg` file 🙂)

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -43,11 +43,11 @@ defmodule Ejabberd.MixProject do
 
   def application do
     [mod: {:ejabberd_app, []},
-     applications: [:idna, :inets, :kernel, :sasl, :ssl, :stdlib, :mix,
-                    :fast_tls, :fast_xml, :fast_yaml, :jose,
+     extra_applications: [:idna, :inets, :kernel, :sasl, :ssl, :stdlib, :mix,
+                    :fast_tls, :fast_xml, :fast_yaml, :jose, :logger,
                     :p1_utils, :stringprep, :syntax_tools, :yconf, :xmpp]
      ++ cond_apps(),
-     included_applications: [:mnesia, :os_mon, :logger,
+     included_applications: [:mnesia, :os_mon,
                              :cache_tab, :eimp, :mqtree, :p1_acme,
                              :p1_oauth2, :pkix]
      ++ cond_included_apps()]


badlop:
Your change seems problematic when running "make dev" in ejabberd.

 What if you move to extra_applications only "logger", instead of all the apps? Something like this, does that work correctly for your user case?

```diff
diff --git a/mix.exs b/mix.exs
index 9093fef9f..ac40e3b1e 100644
--- a/mix.exs
+++ b/mix.exs
@@ -47,7 +47,8 @@ defmodule Ejabberd.MixProject do
                     :fast_tls, :fast_xml, :fast_yaml, :jose,
                     :p1_utils, :stringprep, :syntax_tools, :yconf, :xmpp]
      ++ cond_apps(),
-     included_applications: [:mnesia, :os_mon, :logger,
+     extra_applications: [:logger],
+     included_applications: [:mnesia, :os_mon,
                              :cache_tab, :eimp, :mqtree, :p1_acme,
                              :p1_oauth2, :pkix]
      ++ cond_included_apps()]
```

prefiks:
To use extra_applications, i think we would need to remove from that apps that we specify in deps (as those will be added automatically)

slezakattack:
> What if you move to extra_applications only "logger", instead of all the apps? Something like this, does that work correctly for your user case?

Just tried this and got a warning when I compiled it:
```
Compiling 17 files (.ex)
both :extra_applications and :applications was found in your mix.exs. You most likely want to remove the :applications key, as all applications are derived from your dependencies
Generated ejabberd app
==> ejabberd_mix_release_issue
```

Running `mix release` works too. While it still technically works with badlop's change, mix seems to want either `applications` or `extra_applications` as maybe it's redundant or something.

badlop:
@slezakattack: Good news: I've pushed your commit directly in master, and followed the suggestion from @prefiks , and it works correctly now :-)

slezakattack:
Awesome, thank you so much. Looks much cleaner too 👍

PR #4285, opened 2024-09-26
CI: Add XMPP Interop tests

guusdk:
Modifies the CI build to integrate the GitHub Action-based test runner from the [XMPP Interop Testing project](https://xmpp-interop-testing.github.io). This executes additional integration tests that help verify ejabberd's compliance with XMPP specifications. To date, approximately 400 tests are included.

In this commit, none of the tests are excluded, which likely results in false positives. Therefor, the 'continue-on-error' flag is set, which intends to prevent the 'build from failing' even when one or more of the interop tests fail. This should ideally be removed in a later commit, after the configuration has been fine-tuned toward. If you choose to accept this PR, I would advise to iteratively evaluate the test responses, and for failing tests either adjust ejabberd's functionality (in case an actual bug was identifier), or to add a failing test to the configurable list of test exclusions.

I'm far from an expert in the ejabberd continuous integration setup. There might be more appropriate ways to integrate these tests.

I've rather arbitrarily picked the server that is build and started in the 'Development Release' step as the server that is targeted by these tests. Perhaps the 'Production Release' server is more appropriate?

I'm unsure if the splitting up of the "Check Development Release" step, in distinct 'start' and 'stop' steps (in between the new tests are executed) is appropriate. Does it affect the original purpose of this step?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,7 @@ jobs:
         grep -q "is started in" $RE/logs/ejabberd.log
 
     - name: Run XMPP Interoperability Tests against CI server.
+      if: matrix.otp == '26'
       continue-on-error: true
       uses: XMPP-Interop-Testing/xmpp-interop-tests-action@v1.4.0
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,12 +147,28 @@ jobs:
         cat $RE/logs/ejabberd.log
         grep -q "is stopped in" $RE/logs/ejabberd.log
 
-    - name: Check Development Release
+    - name: Start Development Release
       run: |
         make dev
         RE=_build/dev/rel/ejabberd
+        sed -i 's/starttls_required: true/starttls_required: false/g'  $RE/conf/ejabberd.yml
         $RE/bin/ejabberdctl start
         $RE/bin/ejabberdctl started
+        $RE/bin/ejabberdctl register admin localhost admin
+        grep -q "is started in" $RE/logs/ejabberd.log
+
+    - name: Run XMPP Interoperability Tests against CI server.
+      continue-on-error: true
+      uses: XMPP-Interop-Testing/xmpp-interop-tests-action@v1.4.0
+      with:
+        domain: 'localhost'
+        adminAccountUsername: 'admin'
+        adminAccountPassword: 'admin'
+
+    - name: Stop Development Release
+      if: always()
+      run: |
+        RE=_build/dev/rel/ejabberd
         $RE/bin/ejabberdctl stop
         $RE/bin/ejabberdctl stopped
         cat $RE/logs/ejabberd.log


guusdk:
As your build uses a matrix, the tests run multiple times. I see that the logs that are attached as artifacts to the build only include one set. I'm assuming that the attachments of each run 'overwrite' each-other. That's something that should be improved.

image

guusdk:
I'm unsure why the CI / Tests step is now failing. It wasn't in an earlier version of this PR. If the failure is related to the changes in this PR, then I'm unsure how to resolve them. Help is very much welcome.

badlop:
It seems unrelated, don't worry. In fact, it fails in a run where the PR has no participation.

badlop:
>  add a failing test to the configurable list of test exclusions.

Ok, I've identified the 12 protocols that right now fail with ejabberd. I'll add them later to exclusion and disable the temporary option continue-on-error.

> I've rather arbitrarily picked the server that is build and started in the 'Development Release'

Looks great. If it were necessary in the future, it would be moved to Production, or to a standalone job or workflow.

> I'm unsure if the splitting up of the "Check Development Release" step, in distinct 'start' and 'stop' steps (in between the new tests are executed) is appropriate. Does it affect the original purpose of this step?

Those steps were just to test the release can start and stop correctly. Adding interop tests there looks perfect :)

> For now, I'll limit the execution of these tests to just one build in the matrix

Right, it seems running with erlang 26 release is enough (that's the latest stable, and the one used for the installers and the containers).

In summary, I think the PR is ready for merge.

Once thing that would be great to update in the action itself is this warning:

> Tests (26)
> The following actions use a deprecated Node.js version and will be forced to run on node20: dsaltares/fetch-gh-release-asset@1.1.1. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

guusdk:
Thanks for the feedback! I've raised the issue with the deprecated Node.js version as https://github.com/XMPP-Interop-Testing/xmpp-interop-tests-action/issues/17

guusdk:
We have [documentation](https://xmpp-interop-testing.github.io/documentation/diagnose-test-failures) that helps to identify the cause of test failures.

The test that fails is reporting that it (unexpectedly) received a stream error. This can be seen in the (std-out) output of the test, that contains this stack trace:

```
Oct 16, 2024 11:52:46 AM org.igniterealtime.smack.inttest.SmackIntegrationTestFramework runConcreteTest
SEVERE: LowLevelRosterIntegrationTest.testPresenceEventListenersOffline (LowLevel, tcp) Failed
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework$LowLevelTestMethod.invoke(SmackIntegrationTestFramework.java:970)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.lambda$invokeLowLevel$2(SmackIntegrationTestFramework.java:624)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.runConcreteTest(SmackIntegrationTestFramework.java:555)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework$PreparedTest.run(SmackIntegrationTestFramework.java:763)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.runTests(SmackIntegrationTestFramework.java:543)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.run(SmackIntegrationTestFramework.java:276)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.main(SmackIntegrationTestFramework.java:114)
Caused by: org.jivesoftware.smack.XMPPException$StreamErrorException: invalid-xml You can read more about the meaning of this stream error at http://xmpp.org/rfcs/rfc6120.html#streams-error-conditions
<stream:error><invalid-xml xmlns='urn:ietf:params:xml:ns:xmpp-streams'/></stream:error>
	at org.jivesoftware.smack.tcp.XMPPTCPConnection$PacketReader.parsePackets(XMPPTCPConnection.java:1002)
	at org.jivesoftware.smack.tcp.XMPPTCPConnection$PacketReader$1.run(XMPPTCPConnection.java:953)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

All XMPP output is stored in separate files, which are attached as an artifact to the Github Action Run. For this test, the file name is `LowLevelRosterIntegrationTest.testPresenceEventListenersOffline (LowLevel, tcp).log`

The file ends with receiving the stream error (in the last four lines) on connection number 10. From the same file, we can see this interaction (I've modified the file to exclude data from other connections for readability):

```
11:52:46.572 SENT (10): 
<presence id='C28T9-8' type='unavailable'/>
<r xmlns='urn:xmpp:sm:3'/>
<a xmlns='urn:xmpp:sm:3' h='15'/>
</stream:stream>
11:52:46.573 RECV (10): 
<iq to='sinttest-zgacm-10@localhost/129355925936844079563330' from='sinttest-zgacm-10@localhost' type='set' id='push8563760907716720775'>
  <query ver='1db0fac9a362de677f88cb8a38917829b244c8e3' xmlns='jabber:iq:roster'>
    <item subscription='both' name='Con One' jid='sinttest-zgacm-7@localhost'/>
  </query>
</iq>
<a xmlns='urn:xmpp:sm:3' h='16'/>
11:52:46.573 RECV (10): 
<r xmlns='urn:xmpp:sm:3'/>
11:52:46.573 RECV (10): 
<a h='15' xmlns='urn:xmpp:sm:3'/>
11:52:46.675 RECV (10): 
</stream:stream>
11:52:46.675 XMPPConnection closed (XMPPTCPConnection[sinttest-zgacm-10@localhost/129355925936844079563330] (10))
11:52:46.676 SENT (10): 
<iq to='sinttest-zgacm-10@localhost' id='push8563760907716720775' type='result'>
</iq>
11:52:46.677 SENT (10): 
<stream:stream xmlns='jabber:client' to='localhost' xmlns:stream='http://etherx.jabber.org/streams' version='1.0' from='sinttest-zgacm-10@localhost' xml:lang='en'>
11:52:46.677 RECV (10): ?xml version='1.0'?>
<stream:stream id='13374172780783452347' ve
11:52:46.677 RECV (10): rsion='1.0' xml:lang='en' xmlns:stream='http://etherx.jabber.org/streams' xmlns='jabber:client'>
<stream:error>
  <invalid-xml xmlns='urn:ietf:params:xml:ns:xmpp-streams'/>
</stream:error>
</stream:stream>
```

I think what is happening here:
1. the client closes the stream with `</stream:stream>`
2. the server pushes an IQ request, and also closes the stream with a `</stream:stream>`
3. the client attempts to respond to the IQ request

That sounds like a bug on our end.

guusdk:
I've raised https://igniterealtime.atlassian.net/browse/SMACK-954 to track this issue.

Neustradamus:
Thanks for this improvement!

badlop:
> The test suite has not silently changed.

Ah, that's great :)

> What is likely going on is that either the test implementation or the server implementation is not being consistent. 

Aha, ok.

As I can see, that test case fails regularly in ejabberd's github actions, and I don't know how to disable that specific test case. I guess I'll add `continue-on-error: true` again until the bug is fixed, or resolved, or there's some way to run the already successful tests without giving false positives. I'll keep an eye in the Smack ticket too, thanks!

PR #4306, opened 2024-11-09
Makefile.in: introduce INSTALL_NOT_VERSIONED_DEP

catap:
Some distributives installs all dependencies as `/usr/local/lib/ejabberd/xxx` and adding `-1.2.3` to each of them just make things more noisy.

Here, I suggested a trivial changes with undocumented variable which maintaner may discover and enjoy, when he need it.

   description.

   them as separate PRs.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -277,7 +277,11 @@ indent:
 
 JOIN_PATHS=$(if $(wordlist 2,1000,$(1)),$(firstword $(1))/$(call JOIN_PATHS,$(wordlist 2,1000,$(1))),$(1))
 
+ifdef INSTALL_NOT_VERSIONED_DEP
+VERSIONED_DEP=$(1)
+else
 VERSIONED_DEP=$(if $(DEP_$(1)_VERSION),$(DEP_$(1)_VERSION),$(1))
+endif
 
 DEPIX:=$(words $(subst /, ,$(DEPSDIR)))
 LIBIX:=$(shell expr "$(DEPIX)" + 2)


badlop:
> Some distributives installs all dependencies as /usr/local/lib/ejabberd/xxx

Which ones? I checked Debian, ArchLinux, Alpine, all them follow that convention in their packages for ejabberd, and for some other packages too.

> adding -1.2.3 to each of them just make things more noisy.

Binary files should be versioned in some way, right? Either in the file name or in the directory name, to ensure the correct version of the binary is being used, and not just whatever may be installed in the system.

In this case, ejabberd's "make install" follows the directory structure specified in the Erlang/OTP documentation, relevant links:

- https://www.erlang.org/doc/system/release_structure#creating-a-release-package
- https://www.erlang.org/doc/system/applications#directory-structure-for-a-released-system
- https://adoptingerlang.org/docs/production/releases/

> maintaner may discover and enjoy, when he need it.

What specific maintainer case are you addressing?

catap:
> > Some distributives installs all dependencies as /usr/local/lib/ejabberd/xxx
> 
> Which ones? I checked Debian, ArchLinux, Alpine, all them follow that convention in their packages for ejabberd, and for some other packages too.

For example FreeBSD: https://www.freshports.org/net-im/ejabberd

> > maintaner may discover and enjoy, when he need it.
> 
> What specific maintainer case are you addressing?

myself as potential maintainer maintainer for OpenBSD port.

catap:
> > adding -1.2.3 to each of them just make things more noisy.
> 
> Binary files should be versioned in some way, right? Either in the file name or in the directory name, to ensure the correct version of the binary is being used, and not just whatever may be installed in the system.
> 

If you install everything by OS package managment system under `/usr/local/lib/ejabberd` I don't see any reason to version binary files here.

In the case of update the package managment system will take care of it. Thus, it also impossible install two version of ejabberd here.

Have I missed something?

catap:
Probably it only increase complexity, let close it.

PR #4323, opened 2024-12-03
Enable allow_unencrypted_sasl2 on websockets

mzealey:
https://github.com/processone/ejabberd/commit/47232838 added the allow_unencrypted_sasl2 option, but this was not added to websockets.


weiss:
Thanks! 👍🏻

PR #4327, opened 2024-12-12
Kill orphan `epmd` after `ejabberdctl ping`

catap:
Here a trivial changes that cleanup an orphan `epmd` after `ejabberdctl ping` which migth be used inside some scripts to confirm that ejabberd is stop.

### DIFF
--- a/ejabberdctl.template
+++ b/ejabberdctl.template
@@ -389,6 +389,7 @@ case $1 in
                  -eval 'net_kernel:connect_node('"'$PEER'"')' \
                  -eval 'io:format("~p~n",[net_adm:ping('"'$PEER'"')])' \
                  -s erlang halt -output text
+        stop_epmd
         ;;
     started)
         set_dist_client


badlop:
> ejabberdctl ping which migth be used inside some scripts to confirm that ejabberd is stop.

"confirm that ejabberd is stop"? there's exactly a command to do that, `ejabberdctl stopped` :)

On the other hand, `ping` is used to confirm that ejabberd is running.

Why do you use `ping` instead of `stopped` for that purpose? Maybe it's possible to improve `stopped`.

badlop:
I'm sorry, I don't understand what benefit `ping` provides that `stopped` lacks for checking if ejabberd is stopped.

I'm trying to understand the reason for using one method instead of the other, maybe there is some change that would improve `stopped` usefulness.

On the other hand, you may want to consider dropping epmd completely: did you know ejabberd can start and be administer without using epmd at all? Just enable `ERL_DIST_PORT`, see https://github.com/processone/ejabberd/commit/30f1e284681db482ef78da6d865ccc79120e4ca3

> in my rc script for OpenBSD

Do you mean this is your personal script? Or are you preparing changes for the OpenBSD upstream pakage?

catap:
@badlop let me rephase my phrase.

The rc script for OpenBSD looks like:
```
rc_start() {
	rc_exec "${daemon} ${daemon_flags} start \
		&& ${daemon} ${daemon_flags} started"
}

rc_check() {
	rc_exec "${daemon} ${daemon_flags} ping | grep -q pong"
}

rc_reload() {
	rc_exec "${daemon} ${daemon_flags} reload_config"
}

rc_stop() {
	rc_exec "${daemon} ${daemon_flags} stop \
		&& ${daemon} ${daemon_flags} stopped"
}
```

When someone calls `rcctl start ejabberd` it calls: `rc_start` and after that `rc_check`. The same happens for `rcctl stop ejabberd`: `rc_stop` and after that `rc_check`.

The rc system relays on `rc_check` to confirm that daemon really started or stopped.

Current script works well, with one exception it leaves orphan `epmd` after `rcctl stop ejabberd`.

badlop:
Aha, now I understand the situation thanks!

Your script uses `stopped` to check it was stopped, and `ping` to double check this and any other start/stop operation. As any of those commands start `epmd`, then you would like to stop `epmd` after calling `ping`.

Is it reasonable to assume that some other people, instead of `ping`, may use other command like `status`, and then they will also request to call `stop_epmd` after whatever command they picked? Following that trend, we would call stop_epmd after each single call...

Alternatively, if you are worried about leaving epmd running, you could try my previous suggestion: don't use epmd at all:

> On the other hand, you may want to consider dropping epmd completely: did you know ejabberd can be started and administered without using epmd at all? Just enable ERL_DIST_PORT, see https://github.com/processone/ejabberd/commit/30f1e284681db482ef78da6d865ccc79120e4ca3

If you setup the option `ERL_DIST_PORT` in `ejabberdctl.cfg`, epmd will not be started at all.  Of course all the ejabberdctl features work perfectly, using that port number instead of calling epmd.

catap:
> Is it reasonable to assume that some other people, instead of `ping`, may use other command like `status`, and then they will also request to call `stop_epmd` after whatever command they picked? Following that trend, we would call stop_epmd after each single call...
> 

What seems resanable, because each command my run `epmd` if it needed.

> Alternatively, if you are worried about leaving epmd running, you could try my previous suggestion: don't use epmd at all:
> 
> > On the other hand, you may want to consider dropping epmd completely: did you know ejabberd can be started and administered without using epmd at all? Just enable ERL_DIST_PORT, see [30f1e28](https://github.com/processone/ejabberd/commit/30f1e284681db482ef78da6d865ccc79120e4ca3)
> 
> If you setup the option `ERL_DIST_PORT` in `ejabberdctl.cfg`, epmd will not be started at all. Of course all the ejabberdctl features work perfectly, using that port number instead of calling epmd.

Yep, this is current solution. But I've switched to it not because I'm worried about running `epmd`.

Here different point: epmd may be run as system services, and `stopped` may killls it, if ejabberd the last application which uses it. To avoid this, I've migrated everything to non-epmd setup.

BTW, feel free to close this PR if you think that here no need for this changes.

badlop:
Aha, right. epmdlessness in theory has mostly benefits

ERL_DIST_PORT is supported by ejabberd source code, binary installers and `ejabberd` container image. It isn't enabled by default in any of them, but maybe it makes sense in the future to enable it by default...

If you find any problem with that feature, don't hesitate to report it, so it can get improved and battle-tested.

I'll close this PR, thanks!

badlop:
I've added a mention to ERL_DIST_PORT in the epmd section
https://docs.ejabberd.im/admin/guide/security/#epmd

catap:
Thanks!

FYI: this is enabled out of the box in OpenBSD ejabberd.

PR #4356, opened 2025-02-25
ref: c2s_handle_bind hook

pouriya:


pouriya:
ping

prefiks:
May i ask for what you need this? I don't see anything wrong with that, so i will merge it.

pouriya:
I need to get device information (version, platform, push id, etc) from client and based on this information enable/disable some features before their first presence (being online) or simply prevent some versions from being online.
Since we can change client resource from server side, I collect this info from resource and remove them from it and give it back to the client (to prevent receiving large `from` attribute in stanzas).

I could receive this info using other xeps or by creating a new xep but it adds another new step to logins and maybe new client code to receive/respond to IQs. But with custom resource, the client only needs to construct the resource handle resource binding error.

PR #4364, opened 2025-03-28
ejabberdctl: add new `kick_users host` command

marc0s:
This new command accepts a host name and will kick all user sessions found for that given host. The result is the number of kicked sessions.


badlop:
I made two small changes in commit https://github.com/processone/ejabberd/commit/17b605a32befcfb9e880922aacdcc58530da4f11

PR #4383, opened 2025-05-09
feat: fire new `push_send_notification` hook

marc0s:
Do not route XEP-357 IQ notification directly but run it through registered hooks that will ultimately decide whether or not send it and, if needed, customize the notification contents.

Hooks can return:

  - a modified IQ packet, as per their business logic
  - the atom `drop`, to effectively cancel the push notification

The idea behind this modification is to allow module developers to create modules that can customize the contents of the emitted push notifications. In our specific case, we use this hook to inspect the full message responsible of triggering the push notification and add some custom fields to the XEP-0357 IQ. Moreover, the return value of `drop` will skip sending the push notification at all if needed.


marc0s:
I'm open for suggestions on the `Hook` name, if any other fits better.

I'd also like to know if any other developer is using Emacs and `elp`. I've been having some trouble matching the indentation of the original file's revision (and it still doesn't fully match it, but `elvis` didn't complain). Any tip is welcome.

marc0s:
Any feedback on this one? Should I gather there's no interest in it? :slightly_smiling_face:

weiss:
> Should I gather there's no interest in it? 🙂

No I think the hook makes sense, will have a look this week.  Thanks for the ping.

weiss:
Thanks, and sorry for the delay.

PR #4408, opened 2025-07-05
mod_pubsub_serverinfo

sstrigler:
This adds mod_pubsub_serverinfo from ejabberd-contrib to ejabberd core.

Todo:
 - [ ] add tests


prefiks:
Hello, i added .spec from this to xmpp repository: https://github.com/processone/xmpp/commit/fbaea1bfb4244e4287fe8449144e8e72d9a842d1, as was requested.

sstrigler:
> Hello, i added .spec from this to xmpp repository: [processone/xmpp@fbaea1b](https://github.com/processone/xmpp/commit/fbaea1bfb4244e4287fe8449144e8e72d9a842d1), as was requested.

Is it by intention you‘ve also added something about unified push and removed muc_subscriptions from the header file?

prefiks:
That was intentional, those files are generated, and generator output is not 100% reproducible (for example muc_subscription was not removed - it was just moved further in a file), and unifiedpush.erl which is also generated, was just not added to git repo in last commit, should split that to separate commit probably, but noticed it only i pushed this change.

sstrigler:
> That was intentional, those files are generated, and generator output is not 100% reproducible (for example muc_subscription was not removed - it was just moved further in a file), and unifiedpush.erl which is also generated, was just not added to git repo in last commit, should split that to separate commit probably, but noticed it only i pushed this change.

Thanks for clarification, just wanted to make sure. 

That said, your tests are failing.

PR #4412, opened 2025-07-11
fix: unsubscribe users from members-only rooms when expelled

### LINKED ISSUE #4410: MucSub: revoking membership in members-only room should unsubscribe the user
## Environment

- ejabberd version: 24.07
- Erlang version: Erlang (SMP,ASYNC_THREADS) (BEAM) emulator version 14.2.1

## Bug description

Given a members-only room with its members (affiliation=member) subscribed to it via MucSub, when an owner revokes the membership of one of the members (affiliation=none), the MucSub subscription remains.

This has, at least, two problems:
* the user will still receive messages via the subscription even if it's supposed to be a non-member now, and
* if the user is re-invited to the very same MUC, the persisted (PostgreSQL in my case) state of the room is wrong, because the affiliation of the re-invited user is not reflected in there (I suspect it's due to https://github.com/processone/ejabberd/commit/4a53d4cb56ce677a655d88ea7b818e9666fe40c5 and the state in the DB is not updated because the code thinking it's already a member) 

I'll try to provide a PR if the current behaviour is confirmed to be wrong.

marc0s:
Fixes #4410

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -3221,6 +3221,7 @@ process_item_change(Item, SD, UJID) ->
                             true ->
                                 send_kickban_presence(UJID, JID, Reason, 321, none, SD),
                                 maybe_send_affiliation(JID, none, SD),
+                                unsubscribe_from_room(JID, SD),
                                 SD1 = set_affiliation(JID, none, SD),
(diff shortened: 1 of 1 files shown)


marc0s:
Fixed indentation with `make format`

PR #4416, opened 2025-07-14
Accepting a proxied connection only in debug log

ChaosKid42:
I believe accepting a proxied connection should not be in the log at level INFO. Level DEBUG would be right.

Reason: 

1. other connections are also in DEBUG-Level and not in INFO
2. INFO-level log is cluttered with these log-messages because the reverse-proxy (e.g. haproxy) might check the availability of the ejabberd-backend on short interval (e.g. every 30 seconds or so). Every check generates a log entry on level INFO.

### DIFF
--- a/src/ejabberd_listener.erl
+++ b/src/ejabberd_listener.erl
@@ -341,7 +341,7 @@ accept(ListenSocket, Module, State, Sup, Interval, Proxy, Arity) ->
 				       gen_tcp:close(Socket),
 				       none
 			       end,
-		    ?INFO_MSG("(~p) Accepted proxied connection ~ts -> ~ts",
+		    ?DEBUG("(~p) Accepted proxied connection ~ts -> ~ts",
 			      [Receiver,
 			       ejabberd_config:may_hide_data(
 				 format_endpoint({PPort, PAddr, tcp})),


prefiks:
Non-proxied connections also use info loglevel (https://github.com/processone/ejabberd/blob/master/src/ejabberd_listener.erl#L363], only debug level entries are for connections that ejabberdctl uses to execute commands.
So if we decide to lower level of those messages, we should also do that in that other place.
@badlop what do you think about doing that?

weiss:
FWIW, I'd be against such a change.

`info` level logs connections and XMPP sessions.  That allows administrators to track down issues reported by users or remote admins.  Therefore, you may want those messages during regular operation, not just for debugging ejabberd.

If you have no need for this, you should be using the `warning` level.  If something in between is needed, we may have to thing about that.  But we shouldn't hide connection logging behind `debug` log level, IMO.

badlop:
Looking at [how Erlang](https://www.erlang.org/doc/apps/kernel/logger_chapter#log-level) and [Elixir](https://hexdocs.pm/logger/1.18.4/Logger.html#module-levels) describe the log levels, I would say that `info` is the proper loglevel for that case.

Of course, depending on the configuration and conditions, and what the administrator knows about the current setup, what wants to get informed and alerted... it is understandable that sometimes some admin is not interested at all in some kind of log messages.

Erlang allows to fine-tune log levels per module using [set_module_level](https://www.erlang.org/doc/apps/kernel/logger.html#set_module_level/2). One example usage is the ejabberd toplevel option [log_modules_fully](https://docs.ejabberd.im/admin/configuration/toplevel/#log_modules_fully)

In your case, you can try executing this in your running ejabberd node:
```erlang
logger:set_module_level([ejabberd_listener], warning).
```

If that satisfies you and want to execute this everytime ejabberd starts, a simple and tricky solution is to edit `$HOME/.erlang` and add a line like:
```erlang
timer:apply_after(5000, logger, set_module_level, [[ejabberd_listener], warning]).
```

If this were a common trend, then anothe solution would be to write a small module (maybe in ejabberd-contrib) to manage those fine-tuned loglevel configurations.

ChaosKid42:
Thank you for your detailed explanation. I'll try out your suggestions. Should be good enough for my use case.

PR #4438, opened 2025-08-18
Update Dockerfile

paulo-roger:
Dockerfile: add Python & psycopg2 support for external auth scripts

This change installs python3, pip, and required build deps in both build and runtime stages, and globally installs bcrypt and psycopg2.

Motivation:
- ejabberd supports external authentication scripts written in Python.
- These scripts often require bcrypt (password hashing) and psycopg2 (PostgreSQL driver).
- Currently, users need to extend the official image to add these dependencies manually.

With this change, the official Docker image can run Python auth scripts out of the box.

## Summary

This PR updates the Dockerfile to include Python support and
common dependencies needed for ejabberd authentication scripts.

### Changes
- Install `python3`, `py3-pip`, `python3-dev`, `postgresql-dev`,
  and build tools via `apk`.
- Install Python packages `bcrypt` and `psycopg2` globally.

### Motivation
Many ejabberd deployments rely on custom external authentication
scripts, frequently written in Python. A common pattern is to use:

- `bcrypt` for password hashing
- `psycopg2` for PostgreSQL access

At present, the official Docker image does not include these
dependencies, forcing users to extend the Dockerfile themselves.

### Benefits
- Provides out-of-the-box support for Python authentication scripts.
- Reduces friction for users deploying ejabberd with PostgreSQL and bcrypt-based password verification.
- Makes the image more immediately useful in production without customization.

### Notes
- The additions are lightweight (~a few MB).
- Maintains compatibility with existing builds.
- If the project prefers minimal images, an alternative approach
  would be to publish an additional `-with-python` variant.

---

### DIFF
--- a/.github/container/Dockerfile
+++ b/.github/container/Dockerfile
@@ -19,6 +19,11 @@ RUN go install -v \
 FROM docker.io/erlang:${OTP_VSN}-alpine AS ejabberd
 
 RUN apk -U add --no-cache \
+        python3 \
+        py3-pip \
+        build-base \
+        postgresql-dev \
+        python3-dev \
         autoconf \
         automake \
         bash \
@@ -36,6 +41,9 @@ RUN apk -U add --no-cache \
         sqlite-dev \
         yaml-dev \
         zlib-dev
+
+# Install Python packages globally
+RUN pip3 install --no-cache-dir --break-system-packages bcrypt psycopg2
 
 ARG ELIXIR_VSN
 RUN wget -O - https://github.com/elixir-lang/elixir/archive/v$ELIXIR_VSN.tar.gz \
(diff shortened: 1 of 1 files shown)


badlop:
This opens a few questions:

The external authentication script can be written in any language, right?

In any compiled language like C, Cobol, and also any interpreted languages like Perl, Python, Tcl, Ruby, JavaScript, Lisp, Elixir, etc, right?

And people may use any of those languages for writting their authentication scripts, right?

And, in order to reduce friction for that people, the compilers and interpreters of all the existing languages should be included, right?

But maybe not all languages are nowadays really used, so they don't need to be included. Which ones to include in the image, and which ones to discard as unpopular/irrelevant/obsolete? The ones you use? The ones I use? The most popular nowadays? The most popular according to what statistics? The ones that somebody request (like Python in this case)?

The current solution is to provide a small image for everybody, and let people decide freely which language they use, and adapt the image accordingly.

> If the project prefers minimal images, an alternative approach would be to publish an additional -with-python variant.

Interesting possibility, because I remember the CAPTCHA feature also requires some additional libraries not included in the current image. Maybe there could be an "extended" image with the Python you propose, and any other language other people may request, and CAPTCHA support libraries, etc.

PR #4443, opened 2025-08-25
CI: bump XMPP-Interop-Testing/xmpp-interop-tests-action

guusdk:
Updates this GitHub Action that's used to execute XMPP-based interop tests from v1.6.0 to v1.6.1.

This is a bugfix release that should increase the stability / predictability of test execution.

A notable change is that the file structure in which XMPP stanzas are generated (which is provided as debug output) has changed. They are still stored in the directory denoted by the logDir argument, but the file structure in that directory has changed somewhat.


guusdk:
@coveralls I changed exactly one character. That hasn't changed test coverage. I think you need better support for rounding. :)

PR #4475, opened 2025-10-29
fix nasty ssl warnings

sstrigler:
The way those SSL warning messages are designed today tend to clog up log files without delivering worthwhile information that would enable an admin to take some action. That’s why I’d demote c2s warnings to info msgs while adding source information to s2s_in warnings.


prefiks:
Makes sense, let's merge.

PR #4476, opened 2025-10-29
fix preload_rooms in case of SQL-DB

ChaosKid42:
I found that preloading of rooms does only work in case you use mnesia as database. It does not work in case you use any SQL database. This PR seems to fix it.


prefiks:
Yes, this seems to be full fix for this issue, i will merge it.

PR #4512, opened 2026-01-19
Roster Subscription Pre-Approval

sstrigler:
Implementation of RFC 6121 section 3.4: Pre-Approving a Subscription Request https://datatracker.ietf.org/doc/html/rfc6121#section-3.4

To initiate before anything else, a person that wants to smoothly add a new contact for some other person or bot sends
```xml
<presence type='subscribed' to='other@some_host.tld' id='123'/>
```
Then proceeds with regular subscription request. If the other entity replies with their own subscription request, our server will then automatically approve it.

## Changes in SQL schema

A new column `approved` of type boolean is added to `rosterusers`.

```sql
ALTER TABLE `rosterusers` ADD COLUMN `approved` boolean NOT NULL AFTER `subscription`;
```

[Sponsored by NLnet](https://nlnet.nl/project/ejabberd-invites/)

### DIFF
--- a/test/privacy_tests.erl
+++ b/test/privacy_tests.erl
@@ -308,7 +308,15 @@ deny_group_master(Config) ->
     deny_master(Config, {group, Group}).
 
 deny_group_slave(Config) ->
-    deny_slave(Config).
+    %% [FIXME] needed to make https://github.com/processone/ejabberd/pull/4512 work, but unclear why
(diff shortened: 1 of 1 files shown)


prefiks:
@badlop Database stuff looks ok, i will need to update db_serialize after this, but since it doesn't yet have code for migration between db schema version, i will add that after this is merged.

PR #4515, opened 2026-01-26
Fixes for mod_invites

sstrigler:
Collection of little fixes for mod_invites (as they come)

### DIFF
--- a/src/mod_invites.erl
+++ b/src/mod_invites.erl
@@ -43,7 +43,7 @@
 -export([get_local_identity/5, get_local_features/5]).
 
 %% commands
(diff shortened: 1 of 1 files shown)


src/mod_invites.erl
sstrigler: Would have been nicer if we used `timeout()` here instead, but I didn’t know it exists. Then the name of the opt would also have to change. Sth like `expires_in` or so. Thoughts?

prefiks:
I think this is ready for merge right? Or do you need to add some more changes?

sstrigler:
> I think this is ready for merge right? Or do you need to add some more changes?

Good to merge!

prefiks:
Thank you!

PR #4524, opened 2026-02-10
Invites a11y fixes

sstrigler:
Fixes prompted by an accessibility audit funded by NLnet (done by HAN University
of Applied Sciences).


sstrigler:
[HAN Accessibility Report - JWChat.pdf](https://github.com/user-attachments/files/25205360/HAN.Accessibility.Report.-.JWChat.pdf)

sstrigler:
Things are never perfect. Areas to improve are feedback from registration form. Especially for invalid username and weak password. But I think this should be handled at a central place and give users a better idea of what went wrong and how to fix it. Didn't want to put too much effort in this because it would also mean a lot of struggle with translations. So better to do it once and right.

Issue #4534, opened 2026-02-16
mod_invites: fix redirect_url


sstrigler:
Before creating a ticket, please consider if this should fit the [discussion forum](https://github.com/processone/ejabberd/discussions) better.

## Environment

- ejabberd version: 26.01+
- Erlang version: any
- OS: any
- Installed from: any

## Configuration (only if needed): grep -Ev '^$|^\s*#' ejabberd.yml

```yaml
mod_invites: {}
mod_register:
  allow_modules:
    - mod_invites # ??? maybe also mod_register_web?
  register_url: ...
...
```
I.e. semi-open ibr that will redirect to `redirec_url` but allows to use mod_invites account creation tokens

## Errors from error.log/crash.log

Probably none.

## Bug description

If `redirect_url` is set, registration with pre-auth token still get's redirected and as such doesn't work. Instead `redirect_url` should be disregarded if a valid pre-auth token has been sent.

badlop:
I am not able to reproduce this problem, maybe my configuration misses something, or some step?

---

Using ejabberd git master, enabled mod_invites, mod_conversejs, and also configured:

```erlang
  mod_register:
    ip_access: trusted_network
    redirect_url: http://localhost/testtest
```

Steps:
1. in conversejs, the registration form says "To register, visit..."
2. generate an invitation: `ejabberdctl generate_invite localhost`
3. visit the invite url
4. register the account
5. it's correctly registered
6. i can login using conversejs

---

Now I try allowing only mod_invites to register accounts:

```erlang
  mod_register:
    ip_access: trusted_network
    allow_modules:
      - mod_invites
    redirect_url: http://localhost/testtest
```

1. conversejs does not show the registration form, it can only login
2. generate an invitation: `ejabberdctl generate_invite localhost`
3. visit url
4. register account
5. finishes correctly, account is created
6. i can login using conversejs

sstrigler:
User feedback:

> Tue Feb 17 11:47:20 2026 - rain: 
https://upload.jabberworld.info/u/ec27dcda43c1bdb422cd46afa492a43811cee88a/9qez76HXc3/zb2rhhi58wzJwcKP7yiSesJq4awDxJttQgHF9SfYfpU3TmYnB.jpg
[...]

PR #4535, opened 2026-02-18
set meta.pre-auth to skip redirect_url if token validated

sstrigler:
If mod_register has redirect_url configured, we skip using it if a valid token has been presented.

This way you can have two paths into registration: either an invite token or you redirect to a custom web form.

Fixes processone/ejabberd#4534


ravermeister:
Looking forward to see it merged in master🙃

PR #4551, opened 2026-03-23
feat: new hook `muc_affiliation_changed`

marc0s:
This change adds a new `muc_affiliation_changed` hook to `mod_muc`. Whenever an affiliation is changed this hook is fired.

I talked about it briefly in the ejabberd MUC a few weeks ago.

Maybe sending the whole room's state in the callback is overkill and it may be changed to send only the room's JID instead. Feel free to comment on this.

Any feedback will be appreciated.

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -3236,6 +3236,7 @@ process_item_change(UJID) ->
 
 -spec process_item_change(admin_action(), state(), undefined | jid()) -> state() | {error, stanza_error()}.
 process_item_change(Item, SD, UJID) ->
+    ServerHost = SD#state.server_host,
     try case Item of
 	    {JID, affiliation, owner, _} when JID#jid.luser == <<"">> ->
 		%% If the provided JID does not have username,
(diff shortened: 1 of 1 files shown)


marc0s:
Have you had time to review this one? Do you think this is a change worth merging?

badlop:
> Maybe sending the whole room's state in the callback is overkill and it may be changed to send only the room's JID instead. Feel free to comment on this.

I took a brief look at existing hooks in `mod_muc_room.erl`

When many content of the room state is needed, then the whole state is passed, for example in `muc_filter_message`
`get_room_config`, `muc_log_get_url`

However, when only one or two elements of room state are needed, they are passed directly: `join_room`, `muc_log_add`, `muc_log_check_access_log`. This facilitates debugging (simple arguments logged in crashes, simple arguments required for triggering the hook, simple arguments to write handling functions). The only drawback is when somebody wants to use the hook and needs something else from the room state.


For `muc_affiliation_changed`, considering its name, I guess it makes sense to send as arguments:
- room jid
- user jid
- old affiliation
- new affiliation

You propose to add this hook because you need it for some reason. Are those arguments enough, or do you need anything else from room's state?

marc0s:
Thanks for your feedback.

> You propose to add this hook because you need it for some reason. Are those arguments enough, or do you need anything else from room's state?

In our current use case, those are enough. 

Do you think it'd be better for the hook to have a signature like: `muc_affiliation_changed(Room, User, OldAffiliation, NewAffiliation)`? Both `Room` and `User` of type `jid()`, and `OldAffiliation` and `NewAffiliation` of type `'admin' | 'member' | 'none' | 'outcast' | 'owner' | 'undefined'`.

badlop:
> Do you think it'd be better for the hook to have a signature like: `muc_affiliation_changed(Room, User, OldAffiliation, NewAffiliation)`? Both `Room` and `User` of type `jid()`, and `OldAffiliation` and `NewAffiliation` of type `'admin' | 'member' | 'none' | 'outcast' | 'owner' | 'undefined'`.

Yes, I think that looks perfect for a developer using it.

marc0s:
Signature changed and rebased on top of current `master`.

PR #4553, opened 2026-03-28
dl_invites_page_deps.sh fixups

toastal:
Running on NixOS (& GuixOS-curious) I am acutely aware of scripts using `#!/bin/bash` which isn’t on many systems. `#!/bin/sh` is required for POSIX-compliance so can be relied on & the script is already POSIX-compliant for shell scripts (no Bashisms) so there is no reason to invoke a different shell; many script here are already pointing to `#!/bin/sh`.

To check, I also ran `checkbashisms` & `shellcheck` to make sure things were square. No Bashisms were used, but `shellcheck` suggested quoting all variables to be extra safe (separate commit). Additionally, there are some useless semicolons that can be dropped (separate commit). Feel free to cherrypick if/as needed.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.


toastal:
I may have signed the CLA? I am concerned about privacy so I wish to not give my name or company so I left it blank (same as I don’t want to be on MS GitHub, but it’s unreasonable to delete this account until projects move on). If it is a concern, consider this trivial & committed under 0BSD.

badlop:
Thanks for pointing to this! As jQuery is no longer used, now the patch is slightly smaller.

Also shellcheck is already used for some scripts, but this script was not yet being tested; I'll add it later.

PR #4558, opened 2026-04-01
add --enable-bootstrap=no to configure options to bypass download

### LINKED ISSUE #4554: Make fetching mod_invites dependencies optional
**Is your feature request related to a problem? Please describe.**

When trying to build the project in the Nix sandbox (which has no network access), the build now fails during the dependency fetch phase. Right now the choice is using cURL or npm to fetch deps—neither of which can be used.

Building aside & more practically, `mod_invites` can be both disabled or point to another endpoint all together meaning there is already 2 cases where these dependencies won’t be used at all making it an minimum a wasteful download.

**Describe the solution you'd like**

One of

1. make a configure/environment variable that that can this block entirely (great for those not opting into `mod_invites`)
2. detect first that `priv/mod_invites/static/` isn’t empty before trying to run fetchers
3. fully vendor the `*.min.*` files into the project
4. remove the dependencies entirely

4 would be best since what is on that single page should be fairly small/straightforward which would not demand such a dependency chain. Ideally the CSS classes in the code would be semantic so downstream can easily modify or even replace the style sheet entirely.

**Describe alternatives you've considered**

To try to get the update for Nixpkgs, I need to decide on either running IFD for the POSIX script (which currently doesn’t work, see: #4553) or pull in the *entire npm toolchain* just to call install for fetching dependencies (manually fetching is infeasible since y’all added SRIs to the assets which would mean a failed dep update wouldn’t be discovered until it fails at runtime—which is likely why the build moved off of relying on OS updates). Neither option is ideal (simple cURL but IFD is slow, or a massive toolchain), but can be done, however… (the actual issue)

Even once I do that to get the dependencies, `make` must called which will try to run these fetcher regardless anyhow which fail without network access. Which means the makefile will need patching just to use those fetched dependencies. If there were at least an env var to opt out, I could put the files where they are expected to be (not to mention the derivation could now easily opt the user out to reduce the closure size if `mod_invites` is not used), & everything would run just fine.

**Additional context**

### LINKED ISSUE #4555: Remove jQuery
**Is your feature request related to a problem? Please describe.**

Y’all upgraded to Bootstrap 5.x which finally removed jQuery—but forgot to remove the jQuery dependency. Skimming the invite JavaScript file, I don’t immediately see anything written that requires jQuery either (& it’s usually trivial to remove).

**Describe the solution you'd like**

Remove jQuery… save bytes (both server & client) & minimize dependencies.

**Describe alternatives you've considered**

Have users, both on the server & client download a probably-unused dependency.

**Additional context**

sstrigler:
Called the option --enable-bootstrap=no (or --disable-bootstrap fwiiw) because the jquery dependency will also be removed. 

Fixes #4554

Since bootstrap5 doesn't depend on jquery anymore and we've only been using it at one single place, there was no reason to keep it around any longer.

Fixes https://github.com/processone/ejabberd/issues/4555

### DIFF
--- a/priv/mod_invites/base_min.html
+++ b/priv/mod_invites/base_min.html
@@ -23,8 +23,7 @@
     </div>
     {% block qr_code %}{% endblock %}
     {% block extra_scripts %}{% endblock %}
-    <script src="{{ static }}/jquery/jquery.min.js" integrity="sha384-fgGyf7Mo7DURSOMnOy7ed+dkq5Job205Gnzu6QIg0BOHKaqt4D76Dt8VlDCzcMHV"></script>
     <script src="{{ static }}/bootstrap/js/bootstrap.min.js" integrity="sha384-G/EV+4j2dNv+tEPo3++6LCgdCROaejBqfUeNjuKAiuXbjrxilcCdDz6ZAVfHWe1Y"></script>
-    <script src="{{ static }}/invite.js" integrity="sha384-ov8LmXw6nEmT+QvXBRPMol0e90WFX9ZysvQudQ7H+nYDyolH4K/+GKJhq7qmTk4T"></script>
+    <script src="{{ static }}/invite.js" integrity="sha384-9dl7uTP5+QJJfidqZFZB530NVQnp58oBvUdbj6mjFPKPNUWhl85g4kP9BStp0bpv"></script>
   </body>
 </html>
--- a/Makefile.in
+++ b/Makefile.in
@@ -235,7 +235,7 @@ else
 endif
 
 ifeq (@bootstrap@,true)
-invites-deps: priv/mod_invites/static/bootstrap/ priv/mod_invites/static/jquery/
+invites-deps: priv/mod_invites/static/bootstrap/
 
 priv/mod_invites/static/bootstrap/:
 	$(INSTALL_INVITES_DEPS)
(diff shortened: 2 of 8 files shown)


priv/mod_invites/base_min.html
```diff
@@ -23,8 +23,7 @@ <h1 class="card-header">{%block h1 %}{% blocktrans %}Invite to {{ site_name }}{%
     </div>
     {% block qr_code %}{% endblock %}
     {% block extra_scripts %}{% endblock %}
-    <script src="{{ static }}/jquery/jquery.min.js" integrity="sha384-fgGyf7Mo7DURSOMnOy7ed+dkq5Job205Gnzu6QIg0BOHKaqt4D76Dt8VlDCzcMHV"></script>
     <script src="{{ static }}/bootstrap/js/bootstrap.min.js" integrity="sha384-G/EV+4j2dNv+tEPo3++6LCgdCROaejBqfUeNjuKAiuXbjrxilcCdDz6ZAVfHWe1Y"></script>
-    <script src="{{ static }}/invite.js" integrity="sha384-ov8LmXw6nEmT+QvXBRPMol0e90WFX9ZysvQudQ7H+nYDyolH4K/+GKJhq7qmTk4T"></script>
+    <script src="{{ static }}/invite.js" integrity="sha384-9dl7uTP5+QJJfidqZFZB530NVQnp58oBvUdbj6mjFPKPNUWhl85g4kP9BStp0bpv"></script>
```
toastal, line 27: If the `"bootstrap": "^5.3.8"` version remained the same, how/why did the SRI change? Why is this a part of the removing jQuery commit?

…Which is also to say, what is even the point of this? SRI is best left for sources you don’t trust like CDNs. This just make your maintenance harder & makes it impossible for downstream self-hosters to modify it (using a new version, stripping out unused features, recompiling which shim to support more browsers).

priv/mod_invites/howto
```diff
@@ -0,0 +1,7 @@
+# HOWTO
+
+## How to create checksums for included css and js files used in templates
```
toastal: Nit: proper casing says this should be

* CSS
* JS

priv/mod_invites/howto
```diff
@@ -0,0 +1,7 @@
...
+
+```sh
```
toastal: Did you mean `console`? This isn’t a shell script but something ran on the interactive shell.

priv/mod_invites/base_min.html
sstrigler: Please read the diff more carefully. The change applies to `invite.js`, not `bootstrap.min.js`. The latter hasn‘t changed, the former needed to so we can get rid of jQuery.

(I merged two PRs into one, the title doesn’t reflect this)

Adding those integrity checks was done in context of a security audit. I had similar concerns as the ones mentioned by you but eventually managed to convince myself.

In the spirit of „you touch it, you own it“ they will help us, upstream, to identify issues when included files have been tampered with. Be it by distributions or server admins (or whoever). 

It might make your life a little bit harder, but save us from some headaches. I‘m a big fan of making changes explicit and easy to discover.

sstrigler:
Strange CI error. We could consider - now that it‘s possible - to skip including bootstrap with the test builds since there are no client side functional tests. At least not at the moment.

PR #4562, opened 2026-04-10
auth: Use constant-time comparison in SCRAM stored-key verification

jpds:
Replaces == with crypto:hash_equals/2 in is_password_scram_valid/2 to do a constant-time comparison.


badlop:
This improvement is similar to https://github.com/processone/xmpp/pull/111

PR #4594, opened 2026-07-20
Add XEP-0158 SHA-256 hashcash CAPTCHA challenge

pounceandmiss:
Add support for sha256 pow challenge as described in [XEP-0158](https://xmpp.org/extensions/xep-0158.html#challenge-hashcash). Off by default, doesn't affect ocr captcha forms.

### DIFF
--- a/src/ejabberd_captcha.erl
+++ b/src/ejabberd_captcha.erl
@@ -43,6 +43,10 @@
 	 host_up/1, host_down/1,
 	 config_reloaded/0, process_iq/1]).
 
+-ifdef(TEST).
+-export([verify_pow/3, pow_label/1, insert_pow_challenge/3]).
+-endif.
+
 -include_lib("xmpp/include/xmpp.hrl").
 -include("logger.hrl").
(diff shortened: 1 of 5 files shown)


badlop:
Hi! Do you know if there is any client that already supports this feature?

pounceandmiss:
@badlop hi, yes, we have two: [web](https://dev.narayana.im/anotherim/anotherim-web/pulls/3) and [desktop](https://dev.narayana.im/anotherim/anotherim-desktop/pulls/193)

badlop:
Can you rebase your PR in current master? There will be a small conflict, instead of
```
Id = <<(p1_rand:get_string())/binary>>,
```

it is nowadays
```
Id = misc:strong_alphanum_token(),
```

And one doubt I have: maybe the documentation of the captcha_pow option should mention that the option is used only by mod_register, when registering a new account using In-Band Registration.

That clarification ensures people don't expect that POW when they enable captcha in muc room or mod_register_web.

What do you think?

PR #4617, opened 2026-09-16
acl: Avoid quadratic lookup when reloading ACL and access tables

mremond:
load_tab/3 checked each old key with lists:keymember/3 against the new list, which is O(N*M) and becomes slow on configuration reload with many virtual hosts.


mremond:
Measured on the isolated algorithm (7 rules per host, one rule removed on reload), on my desktop:

| Hosts  | Before  | After  |
|--------|---------|--------|
| 1,000  | 281 ms  | 3 ms   |
| 5,000  | 7.2 s   | 13 ms  |
| 10,000 | 30.1 s  | 42 ms  |

prefiks:
Alternative fix commited in 6a38f286e4d9e696a484ded5ba61e8630fb2c6e4

